PR #28, opened 2022-12-07
Basic crawler to identify invalid href

### LINKED ISSUE #26: Generic crawler using Beautiful Soup
Implement a generic and simple crawler using Beautiful Soup that can be extended to meet specific parsing requirements under this project.

This crawler should handle

* absolute and relative paths
* cyclic references
* `http` and `file` based uris

rnllv:
- [x] Able to find invalid href
- [x] Prevent cyclic href traversal
- [x] Check for `file` scheme


crawler/simple_crawler.py
```diff
@@ -0,0 +1,111 @@
...
+        self.unreachable_child_resources = {}
+
+    class Resources:
+        def __init__(self, url, parent_url):
+            m = hashlib.sha256()
+            m.update(bytes(url, 'utf-8'))
+            self.id = m.hexdigest()
+            self.url = url
+            self.parent_url = parent_url
+
```
IanMayo, line 52: these two methods are poorly named, in my opinion @rnllv 

They are called `add`,  but they also return true/false information.  _Maybe_ someone would expect `add` to return a new list length, but not a flag for if the item was already present in the list. This is hidden behaviour.

I _think_ I'd rather not have the methods.

I _think_ it would be neater to have at line 82:
```
  child_resource_already_accessed = resource.id in self.visited_child_resources
  if not child_resource_already_accessed:   
     self.visited.child_resources[resource.id] = resource  
```

rnllv: >This is hidden behaviour.

Coming to think of it, yes, it is indeed hidden behaviour.

I'll incorporate the changes you've suggested.

rnllv: Ohh, I forgot. The method was intentional. I agree naming could be improved.

I need those methods there so that if we decide to take a multi-threaded approach in future (for performance), the body in these methods will become `critical sections` (since the dictionaries are shared)  and can be protected by a `lock` object to prevent race conditions.

rnllv: Isolating the `critical sections` in dedicated methods would be more easy to debug than to put it throughout the crawler.

IanMayo: YAGNI.  If we do go multi-threaded in the future, then let's introduce the complexity at that point.

rnllv: :)
Okay. Let me implement the code you suggested.

rnllv: Suggested review changes pushed. Please check.

IanMayo (approved):

Issue #56, opened 2022-12-13
Some class types get merged cells


IanMayo:
I have a country, and see the list of classes for it.  For two classes, their columns for `Power` and `RR` are merged.  This is making `create_new_class_with_extracted_subcategory` fail.  Will reproduce in mock data.

PR #59, opened 2022-12-14
Generic formatting changes and message cleanup

rnllv:
Generic set of formatting and message changes on top of the fixes that were made previously.

### DIFF
--- a/legacyman_parser/parse_classes_of_country.py
+++ b/legacyman_parser/parse_classes_of_country.py
@@ -57,9 +57,9 @@ def extract_classes_of_country(soup: BeautifulSoup = None, parsed_url: str = Non
     global _class_table_header_is_identified, _current_subtype_id
     _class_table_header_is_identified = False
     _current_subtype_id = None
-    classList = soup.find('div', {"id": "PageLayer"})
-    if (classList):
-        for row in classList.find('table').find_all('tr'):
+    class_list = soup.find('div', {"id": "PageLayer"})
+    if class_list:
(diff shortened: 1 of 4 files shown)


legacyman_parser/parse_classes_of_country.py
```diff
@@ -142,20 +142,21 @@ def create_new_class_with_extracted_subcategory(row: PageElement, country: str,
...
+                                       columns[2].text,
+                                       columns[3].text,
+                                       columns[4].text,
+                                       columns[5].text,
+                                       columns[6].text,
+                                       has_tonal,
+                                       tonal_href))
     else:
-        print('Warning: encounterd class with too few columns', parsed_url)
+        print('\nWarning: encountered class with too few columns. url: {}\n'.format(parsed_url))
```
IanMayo, line 158: @rnllv - with this way of reporting, the "warning" is displayed inline with the other parsing text, and could get missed.  Could you modify it so that it gets added to a list of discrepancies (or several lists of discrepances) that get reported at the end of the import please?

IanMayo (changes requested):
Just the one change, please.

legacyman_parser/parse_classes_of_country.py
rnllv: Yep, handled as part of PR #60 (Issue #56). Please approve this PR, and then #60 should be merged without any conflict.

IanMayo (approved):

PR #63, opened 2022-12-14
41 generate unique ids from type parsers

### LINKED ISSUE #41: Generate unique ids from type parsers
The individual type parsers should generate unique ids for entities at the time of extracting them.

rnllv:
fixes #41 
This also generates compatible mock json to be tested in DataBrower.

image


IanMayo (approved):
Great.

PR #78, opened 2022-12-16
75 tests for known discrepancies

### LINKED ISSUE #75: Tests for known discrepancies
Generate test cases for

- [ ] Country hrefs starting with .. should not be processed.

rnllv:
Tests for parse_region to not include countries starting with '..'

Note: countries starting with '..' will be handled in a separate parser (like parse_special_regions.py)


IanMayo (approved):

IanMayo:
Hello @rnllv - the tests are valid, and I can see they help us out.

But, I'd like us to discuss/consider a different strategy.  I honestly don't know if this alternate strategy is better or not.  Let me explain it.  This PR uses a test to verify a method works as expected.

But, we're going to throw these methods away, eventually.

How about we try another strategy?   We run the full `parse` on the `mock` data, then we test the results - whether things are, or aren't present in the output JSON.

So, when I encounter something "breaking" in the real data, I reproduce it in the "mock" data, and write a (failing) test that highlights the issue.  You then modify the `parse` code to fix the test.

Hmm, at some point, we're going to switch to writing to DITA, but right now I think it may be worth running DITA and JSON exports in parallel - so the range of entities present in the JSON grow with the range of entities in DITA.

I'm not quite sure how the above will suit the thing we're testing here.  It will probably fail.  We can't test for the absence of Britain (`../Britain/`), since we currently import Britain into Europe, too.  That's a silly aspect of the mock data  - so I think it would be for me to modify (fix) the mock data.

What are your thoughts on all that?

rnllv:
Your alternate strategy will definitely complement our normal test strategy. Let's discuss this in a new issue.

PR #80, opened 2022-12-19
Test the generated json file

### LINKED ISSUE #79: Alternate test strategy for LegacyMan Parser
Hello @rnllv - the tests are valid, and I can see they help us out.

But, I'd like us to discuss/consider a different strategy.  I honestly don't know if this alternate strategy is better or not.  Let me explain it.  This PR uses a test to verify a method works as expected.

But, we're going to throw these methods away, eventually.

How about we try another strategy?   We run the full `parse` on the `mock` data, then we test the results - whether things are, or aren't present in the output JSON.

So, when I encounter something "breaking" in the real data, I reproduce it in the "mock" data, and write a (failing) test that highlights the issue.  You then modify the `parse` code to fix the test.

Hmm, at some point, we're going to switch to writing to DITA, but right now I think it may be worth running DITA and JSON exports in parallel - so the range of entities present in the JSON grow with the range of entities in DITA.

I'm not quite sure how the above will suit the thing we're testing here.  It will probably fail.  We can't test for the absence of Britain (`../Britain/`), since we currently import Britain into Europe, too.  That's a silly aspect of the mock data  - so I think it would be for me to modify (fix) the mock data.

What are your thoughts on all that?

_Originally posted by @IanMayo in https://github.com/DeepBlueCLtd/LegacyMan/issues/78#issuecomment-1357702369_

rnllv:
supports #79 

@IanMayo, please review. 

This is the decoupled version of tests I was referring to. Here we can specify test_payloads for mock and real environments. We can even ask clients to come up with their own test payload, that they like to be tested (similar to mock_data_test_payload.json), that can be executed against predefined test cases in parsed_json_test_definitions.py.

Execute the following command from the project root directory to run program. The `test_payload_for_parsed_json` is optional parameter.

```
python -m legacyman_parser.parser data/PlatformData/PD_1.html test_payload_for_parsed_json=legacy_tester/test_payload/mock_data_test_payload.json
```

### DIFF
--- a/legacy_tester/test_payload/mock_data_test_payload.json
+++ b/legacy_tester/test_payload/mock_data_test_payload.json
@@ -1,0 +1,3 @@
+{
+  "phase_one_regions_parse_not_included_countries": ["Britain"]
+}
--- a/legacyman_parser/parser.py
+++ b/legacyman_parser/parser.py
@@ -3,6 +3,7 @@ import sys
 
 from crawler.simple_crawler import SimpleCrawler
 from legacy_publisher import json_publisher
+from legacy_tester.parsed_json_tester import parsed_json_tester
 from legacyman_parser.parse_classes_of_country import extract_classes_of_country, CLASS_COLLECTION, SUBTYPE_COLLECTION, \
     TOO_FEW_PROPERTIES, NON_STANDARD_COUNTRY
 from legacyman_parser.parse_countries import extract_countries_in_region, COUNTRY_COLLECTION, COUNTRY_TABLE_NOT_FOUND
@@ -48,6 +49,15 @@ def parse_from_root():
 
     if path_has_back_slash(arg_url) or path_has_drive_component(arg_url):
         cleansed_url = get_cleansed_path(arg_url)
+
+    # test_payload_for_parsed_json
+    kw_args = dict(arg.split('=') for arg in sys.argv[2:])
+    test_payload_json = None
+    if 'test_payload_for_parsed_json' in kw_args:
+        test_payload_json = kw_args['test_payload_for_parsed_json']
+        if not os.path.isfile(test_payload_json):
+            print("Cannot find json test payload at {}".format(test_payload_json))
+            return
 
     """Parsing Region
     The regions are processed from map"""
@@ -127,13 +137,17 @@ def parse_from_root():
         for no_tonal_table in TONAL_TABLE_NOT_FOUND:
             print("    " + no_tonal_table)
 
-    json_publisher.publish(parsed_regions=REGION_COLLECTION,
-                           parsed_countries=COUNTRY_COLLECTION,
-                           parsed_classes=CLASS_COLLECTION,
-                           parsed_tonals=TONAL_COLLECTION,
-                           parsed_subtypes=SUBTYPE_COLLECTION,
-                           parsed_tonal_types=TONAL_TYPE_COLLECTION,
-                           parsed_tonal_sources=TONAL_SOURCE_COLLECTION)
+    published_json = json_publisher.publish(parsed_regions=REGION_COLLECTION,
+                                            parsed_countries=COUNTRY_COLLECTION,
+                                            parsed_classes=CLASS_COLLECTION,
+                                            parsed_tonals=TONAL_COLLECTION,
+                                            parsed_subtypes=SUBTYPE_COLLECTION,
+                                            parsed_tonal_types=TONAL_TYPE_COLLECTION,
+                                            parsed_tonal_sources=TONAL_SOURCE_COLLECTION)
+
+    print("\n\nTest results:\n")
+    if test_payload_json is not None:
+        parsed_json_tester(published_json, test_payload_json)
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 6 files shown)


IanMayo:
Hello @rnllv - could you give me a little more explanation of how this is working?

What is the purpose of the optional JSON parameter?  Is that published JSON document that is to be tested against?  I _think_ we can just run the tests against the JSON that `publish` has just generated.

Hmmm, _or_ we could introduce a python command that just runs `test` against a named JSON document produced by `parse/publish`.   That will avoid having to re-run the parse/publish, if it's just the tests that are being modified.

rnllv:
>could you give me a little more explanation of how this is working?

Sure. Do you have any specific scenario/test case in mind that you wish to be checked against the parsed JSON document? I'll be able to better explain with that scenario.

>What is the purpose of the optional JSON parameter?

The JSON parameter (optional) references to what I call the "test payload file." If provided, the application will run the parser_tests with that payload against the parsed json/DITA output, else the application will halt immediately after the parsed json/DITA output is generated.

>Is that published JSON document that is to be tested against? 

Yep, it is the published JSON document that will be tested against.


>I think we can just run the tests against the JSON that publish has just generated.

This is exactly what this change is doing. All test cases we currently have/planning to have just test the parse logic given a HTML input. This commit will test the parsed JSON that is generated as part of `publish`


>Hmmm, or we could introduce a python command that just runs test against a named JSON document produced by parse/publish. That will avoid having to re-run the parse/publish, if it's just the tests that are being modified.

I don't believe it matter from where we perform these tests. Right now, this will be called as part of the `publish` step itself, if a valid "test payload" is provided. We can move it to `test` if required. But we'll need to execute the test step manually after the parse step each time.

IanMayo:
> test payload file

Aaah, so it's the expected/unexpected data that is used as an input to the tests, right?

(your other comments all make sense, now I understand the role of the passed-in file).

rnllv:
>it's the expected/unexpected data that is used as an input to the tests

Exactly.

This way we can have one payload maintained by us for mock data. And another payload maintained by you/client for the real data.

rnllv:
@IanMayo, can we merge this now? We'll keep enhancing the test definitions and test payload one issue at a time.

legacy_tester/test_payload/mock_data_test_payload.json
```diff
@@ -0,0 +1,3 @@
+{
+  "phase_one_regions_parse_not_included_countries": ["Britain"]
```
IanMayo, line 2: @rnllv - this data structure looks like a dictionary.  The above item is a named field that contains an array of country names.
But, in the test method, I don't see the named field referred to.  Shouldn't `phase_one_regions_parse_not_included_countries` method refer to the dictionary entry of the same name?

IanMayo:
> @IanMayo, can we merge this now? We'll keep enhancing the test definitions and test payload one issue at a time.

You should take it out of `draft` mode when it's ready for review/merge, please.

legacy_tester/test_payload/mock_data_test_payload.json
rnllv: The array of country names is passed as argument (test_payload) to the method.

IanMayo: Yes, sure.  But. I don't see us referring to the named entity above. I just see us parsing a list of countries.  Once we have several named entities in there, we'll need to address them individually in order to use the correct test data in the relevant test.

rnllv: It's referred in line 15 of parsed_json_tester.py and the corresponding method is executed couple of lines after that.

rnllv: Let me clean it up a bit.

IanMayo: Aah, learning moment.  So, the test code works through the dictionary.  It goes one term at a time, and calls a method which matches the term title (`test_method`), passing the test payload to it.

Is that right?

rnllv: Yep.

legacyman_parser/parser.py
```diff
@@ -49,6 +50,15 @@ def parse_from_root():
     if path_has_back_slash(arg_url) or path_has_drive_component(arg_url):
         cleansed_url = get_cleansed_path(arg_url)
 
+    # test_payload_for_parsed_json
+    kw_args = dict(arg.split('=') for arg in sys.argv[2:])
+    test_payload_json = None
```
IanMayo, line 55: @rnllv - I think we can simplify the next two lines.  We check for the  presence of the keyword, and we try to extract it if it is present. I _guess_ we could just try to extract it, then check that `test_payload_json` is `not None` (like we do later on).

rnllv: Python will [throw](https://stackoverflow.com/a/10116540/11905887) a [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError).

This is one aspect that Java handles better than Python :)

Diff:
image

Result:
image

IanMayo: Sure, ok. Thanks for the teach-in.

IanMayo (approved):

PR #88, opened 2022-12-23
76 spotting correct platform sub type

### LINKED ISSUE #76: Spotting correct platform sub-type
In our parsed data there are lots of instances of blocks of text being interpreted as `Platform sub-type`.  

I've checked the data, and it appears that _real_ sub-type rows have a col-span of `7`, _AND_ a `bgcolor` of `#CCCCCC`

We should update the mock data to match this, modify the parser to exploit this,  _AND_ use unit test to verify that `Link is to French unit_d` (from `Spain.html`) _ISN'T_   in the list of `Platform sub-types`.

rnllv:
fixes #76 
image


IanMayo:
Ok, to test this, could we have modify `Narnia1` to include a test marker at the end of this text snippet?
image

Since it doesn't have `bgcolor`, we should not process it.  So, then write/run test to check that no platform-types contain that marker

rnllv:
Yep, that sounds good. Let me include that.

rnllv:
Done.

Test failure scenario:
image

Please review.

legacy_tester/parsed_json_test_definitions.py
```diff
@@ -20,3 +20,17 @@ def count_of_tonal_remarks_containing_test_string(published_json, test_payload):
...
             return False
     return True
+
+
+def subtypes_should_not_have_string(published_json, test_payload):
+    """Test for issue #76
+    This is to test if the parser has successfully ignored <td> blocks that don't
+    contain attributes pertaining to platform sub type <td>
+    """
+    for test_string in test_payload:
```
IanMayo, line 30: hmm, it looks like `test_payload` for `subtypes_should_not_have_string` doesn't have a `test_string` named property. It looks like it just has the string to match.

rnllv: The data type of `test_payload` is purely contextual. In this case, it contains a list of string each of which should not be present in the extracted sub types just like `Britain1` in `phase_one_regions_parse_not_included_countries`

IanMayo:
> Done.
> 
> Test failure scenario: image
> 
> Please review.

For the screenshot, did you temporarily introduce a "breaking" block of text?  I don't understand why it is reporting as "Composite" being broken.

rnllv:
>For the screenshot, did you temporarily introduce a "breaking" block of text?

Yep, I just wanted to make sure the tests indeed worked, and that could be done only with a proper sub type name.

IanMayo (approved):

IanMayo (approved):

PR #90, opened 2022-12-28
30 parse abbreviations page

### LINKED ISSUE #30: Parse abbreviations page
The legacy content includes an abbreviation table [here](https://deepbluecltd.github.io/LegacyMan/data/QuickLinksData/Abbreviations.html).  We should load all of this data into _some_ structure in the DITA.

As we are publishing to HTML and/or exporting to JSON, we should provide a tooltip containing the full name for tonal sources (most of which are abbreviations).

I think the valid strategy for this is to start at the "real" introduction page (`Introduction/ContentsPage.html`) - then we can deliberately navigate to both the platform data _and_ the abbreviations data.

Hmm, actually - we can hard-code `Introduction/ContentsPage.html` - since that's the path in both the mock data and the real data.  Then we only have to pass the path to the root folder to `legacyman_parser.parser`.

rnllv:
fixes #30


target/json_publication.js
```diff
@@ -313,7 +313,7 @@ var publicationJsonData={
       "unit_id": 4,
       "tonal_type_id": 1,
       "source": 1,
-      "ratio": 45.636,
```
IanMayo: Hello @rnllv - could you `seed` the random parser please, so the randomised rations don't change each time you re-generate the content?

See here: https://www.w3schools.com/python/ref_random_seed.asp

rnllv: >See here: https://www.w3schools.com/python/ref_random_seed.asp

Thanks for sharing, @IanMayo. Will implement it.

IanMayo (changes requested):
New item added to issue (testing assumptions).

We can't test the assumptions here in our unit tests, we need to test against the real data, so the testing has to happen in the actual parsing code, please.

rnllv:
Tests to assumptions added.

These are the results with false data (failing as expected)

image

image

image

IanMayo:
> We can't test the assumptions here in our unit tests, we need to test against the real data, so the testing has to happen in the actual parsing code, please.

I _think_ the way you have implemented it, is that we test our assumptions using unit tests against the mock data.  We can't run our tests against the real data, since it won't contain the `special` strings.  But, it's in the real data that we want to test our assumptions. Well, we potentially want to test the assumptions against mock _and_ real data.

So, to test the assumptions - I think we need to add code to the parsers.  This code won't run actual tests, but will verify assumptions, and provide feedback in the parse summary as necessary.

rnllv:
This is how `assert` errors would look like. It'll stop processing immediately.

image
image

legacyman_parser/parse_abbreviations.py
```diff
@@ -0,0 +1,39 @@
...
+    def __str__(self):
+        return "{}. {} ==> {}".format(self.id, self.abbreviation, self.full_form)
+
+
+def parse_abbreviations(soup: BeautifulSoup = None,
+                    parsed_url: str = None,
+                    parent_url: str = None,
+                    userland_dict: dict = None) -> []:
+    # Assumption: There's going to be only one table in QuickLinksData/Abbreviations.html
+    assert len(soup.find_all('table')) == 1, "InvalidAssumption: There's going to be only one table in " \
```
IanMayo, line 23: I know it's _thorough_ to include the comment before the assert, but I think it would be better to just let the assert message explain what it's doing.  Then we only have to describe the assumption in one place - and there's no chance of the two statements not matching each other.

rnllv: Got it.

IanMayo (approved):

PR #95, opened 2023-01-17
Extract flag and class tonal images

### LINKED ISSUE #48: Extracting Class images
Should the class images array be extracted from tonals page?

rnllv:
fixes #48


rnllv:
I see only one image (flag image) for all non-Britain pages, except for Spain where it's the second.
For Britain (and similar pages), the first image is/should be the flag image.

@IanMayo, I'll implement the flag extraction based on the above, and put in required assert statements to validate our assumptions.

IanMayo:
ok, some conflicts to be resolved (sometime), please @rnllv

rnllv:
@IanMayo, conflict resolved. Please review.

IanMayo:
@rnllv - we've added the `Flags` block towards the end of the JSON.  IMHO it would be neater to make `flag` an optional child within a `country`.  Plus, the `Flags` section looks like it is de-normalising other structures - which introduces some duplication/verbosity.

Is there a good reason for the `Flags` block?

rnllv:
>IMHO it would be neater to make flag an optional child within a country

Yep, it should be. And there are few other changes in the class/unit properties required as mentioned in #100 

My plan was to make all those changes in one issue so that we can refer it again if required.

This issue was just to extract the images and place it (somewhere) in the json.

IanMayo:
> My plan was to make all those changes in one issue so that we can refer it again if required.

Sure, ok

IanMayo (approved):

PR #132, opened 2023-03-31
124 find images

### LINKED ISSUE #124: Challenge in finding flag image
There are a range of images on each page, more than the two we were preparing ourselves for.  Some are used as next/back buttons. Others are company logos, or "new" content.

To find the flag, we'll have to use a strategy something like this:
* find all `img` files
* check that `img.src` starts with `./Content/Images`
* check that `img.src` does not end with one of these `image020.jpg`, `Yellow3.jpg`, `Yellow4.jpg`
* check that `img.src` does not end with `.gif` (since this is an animated gif of the flag in the wind, which is not used)

If there is more than one image remaining, filter for images that match `*flag*.jpg`.

There should be one remaining image.  Once we have list of countries that fail this assert, I can either look for more tests, or tidy the raw data.

IanMayo:
Fixes #124


legacyman_parser/parse_flag_of_country.py
```diff
@@ -50,17 +79,27 @@ def extract_flag_of_ns_country(soup: BeautifulSoup = None, parsed_url: str = Non
...
+    in_content_images = list(filter(in_content_images_filter, images))
+    non_blacklisted = list(filter(not_blacklisted_filter, in_content_images))
+    flag_images = non_blacklisted
+
+    assert len(flag_images) == 1 or len(flag_images) == 2, "InvalidAssumption: There should one or two images in this Graylayer, " \
+        "that of the flag: {}. Found {}".format(
+        parsed_url, flag_images)
     flag_images = soup.find_all('img')
     # Always take the first flag
     flag_image = urljoin(parsed_url, flag_images[0]['src'])
```
rnllv, line 94: @IanMayo, if we're not blacklisting *gif files, then we should be confident that gif files will always be second in the list. Otherwise we'll need to blacklist gif files, and assert for just 1 flag image.

IanMayo: There are actually some animated gifs in the dataset, though I don't think they are ever linked to.  But, some actual flags are `.gif` files.  This PR will help us move forward.  I _may_ end up converting actual flags to `.jpg` - so we can be more consistent in our processing.

rnllv: Okay then.

rnllv (approved):

PR #137, opened 2023-04-05
Filter out  ns countries from countries within region

### LINKED ISSUE #125: non-standard country within region
We define "non-standard" (`n-s`) country as one which has a welcome page that breaks the platforms down by type.

So, for `Britain` we don't start off with a table of platform summaries, broken down by type.  Instead there are a series of images, each of which links to the conventional table.

We recognise Britain as a n-s country since in the map of the world it starts off with a `../` relative path reference, rather than linking directly to a region page in the current folder.

But, we have a n-s country that is within a conventional region.  We can spot that it's a n-s country since it does not contain a `PageLayer` div.  The child pages it links to all have a `href` beginning with `../`.

Will collate demo data.

rnllv:
Fixes #125


IanMayo (approved):
Code looks valid. Look forward to testing it this afternoon.

PR #141, opened 2023-04-05
Handle one-off parsing logic for class attribute extraction from tonals page

### LINKED ISSUE #128: Multiple h1 elements in tonals page
We wish to find the tonals table on a class page.

Our initial way of doing that was to look for the table following a `<h1>Propulsion</h1>` element.

But, that is failing, because in one class there is an unrelated table before the tonals table.

So, we need a new test.  This _looks-like_ a generic pattern:
- find all tables on the page
- extract first `td` from the first `tr`
- check if the `td` has a `colspan`, and check value of the `colspan` is `4`
- if it does, attempt to extract the tonal data from it.

rnllv:
Fixes #128

### DIFF
--- a/legacyman_parser/parse_class_attributes_from_tonals.py
+++ b/legacyman_parser/parse_class_attributes_from_tonals.py
@@ -23,19 +23,20 @@ def extract_class_attributes_from_tonals_page(soup: BeautifulSoup = None, parsed
                                               parent_url: str = None, userland_dict: dict = None) -> []:
     # Find all tables on page
     all_tables = soup.find_all('table')
-    propulsion_system_table = None
-    for potential_propulsion_table in all_tables:
+
+    def four_col_table_filter(tag):
         # Find first table data and its colspan
-        tdata = potential_propulsion_table.find('td')
-        if tdata.get('colspan', None) is None:
-            continue
-        if tdata['colspan'] == '4':
-            propulsion_system_table = potential_propulsion_table
-            break
-    assert propulsion_system_table is not None, "InvalidAssumption: There's always a Tonal Propulsion table with " \
-                                                "colspan 4 "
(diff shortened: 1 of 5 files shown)


legacyman_parser/parse_class_attributes_from_tonals.py
```diff
@@ -21,15 +21,22 @@
...
-                                                       "tonals page {}".format(parsed_url)
-    propulsion_system_table_data = propulsion_h1.find_next('table').find_all('td')
-    assert propulsion_system_table_data[0]['colspan'] == '4', "InvalidAssumption: First row in propulsion table " \
-                                                              "is going to be unit title : {}".format(parsed_url)
-    assert len(propulsion_system_table_data) % 2 == 1, "InvalidAssumption: Propulsion table is in" \
-                                                       " Key, Value format, with header : {}".format(parsed_url)
-
+    # Find all tables on page
+    all_tables = soup.find_all('table')
+    propulsion_system_table = None
```
IanMayo: What? You're not using `filter` functions?, like the ones I only learned about this week?

Here is an example of the use of filters:
https://github.com/DeepBlueCLtd/LegacyMan/pull/139/files#diff-38b58fe72d86c4e4ae1fa963ce1c730732c2ea80dbb66fd4ff7dfe6f79f32903R30

I _think_ they let Python do more of the work.  They also have no _side-effects_. The loop structure modifies `propulsion_system_table`, which is outside the code block

It _would_ be possible to combine the tests I was doing in the filters above - so there was just one filter, but I think the code is cleaner, when it's one test per filter.

Aah yes, it's a more _functional_ style of programming, as described [here](https://realpython.com/python-filter-function/#coding-with-functional-style-in-python)

IanMayo (approved):

PR #145, opened 2023-04-06
133 Set current directory drive name to TARGET_DIRECTORY

### LINKED ISSUE #133: Set the target directory from argument passed to the parser
Right now, the value for `TARGET_DIRECTORY` variable in legacyman_parser/utils/constants.py is hardcoded.
Set it from argument and update README accordingly.

rnllv:
Fixes #133


IanMayo (approved):

legacyman_parser/parser.py
```diff
@@ -52,6 +53,10 @@ def parse_from_root():
         print("Url empty. Exiting")
         return
 
+    # Set target value based on cwd
+    if path_has_drive_component(os.getcwd()):
```
IanMayo, line 57: ok, problem with the next line.  It sets the target directory as `target` in the root folder of the current drive. It should set the target as `target` in the current working directory.

PR #158, opened 2023-04-06
154 commonly detected sources

### LINKED ISSUE #154: Trouble with "Commonly detected sources"
There a couple of files where there is a warning about the accuracy of commonly detected sources.

The warning is highlighted to the user via an asterisk (`*`) before the `Commonly Detected Sources` text.  This is breaking our logic that finds the tonals list.

Normal case:
```
<td><strong>Commonly Detected Sources</strong></td>
```

This case:
```
<td><strong><span>*</span>Commonly Detected Sources</strong></td>
```

IanMayo:
Fixes #154 

Note: have tested it against the target data, as well as our mock data. The solution works for all instances.


legacyman_parser/parse_tonals_of_class.py
```diff
@@ -40,9 +40,17 @@ def extract_tonals_of_class(soup: BeautifulSoup = None, parsed_url: str = None,
...
+    # text, but without looking inside child blocks (recursive=False)
+    strong_blocks = soup.find_all("strong")
+
+    def common_block(tag):
+        return tag.find(string="Commonly Detected Sources", recursive=False)
+    tonal_texts = list(filter(common_block, strong_blocks))
+    if len(tonal_texts) == 1:
+        tonal_text = tonal_texts[0]
+        tonal_table = tonal_text.find_parent("table")
         if tonal_table:
```
rnllv, line 54: Can we replace the above changes with a simple functional check as below?

def td_has_commonly_detected_sources(tag):
.      return tag.name == 'td' and "Commonly Detected Sources" in tag.text

tonal_header = soup.find(td_has_commonly_detected_sources)

IanMayo: I don't think so, because `tag.text` wouldn't match with `Commonly detected sources`.  I think it's returning empty string, as the empty element between `<td>` and `<strong>`.

image

Aah, but this may work:
```
return tag.name == 'strong' and "Commonly Detected Sources" in tag.text
```

I can try that this afternoon

IanMayo: Aah, no. The above pattern does not match this:
```
<strong><span>*</span>Commonly detected sources</strong>
```

IanMayo:
Am going to merge this, so I can test the implemented version against real data.

legacyman_parser/parse_tonals_of_class.py
rnllv: Ok

PR #163, opened 2023-04-11
159 - don't extract blacklisted images

### LINKED ISSUE #159: Exclude some class images
When we are extracting the flags, we do some filtering to exclude some images that are presentational/layout related.

We should do the same for `parse_images_of_class`.

Look at the extracted images, and come up with some patterns for what should be deleted.

To include:
* `check_db.gif`
* `image020.jpg`
* `prev_db.GIF`
* `Rtn2Map_db.gif`

We _may_ also need some fancy processing to exclude flags at a later date. We'll see.

IanMayo:
Fixes #159


IanMayo:
Merging, so I can test against real data.

PR #164, opened 2023-04-11
153 alternate way of building up path constants

### LINKED ISSUE #153: Target directory should always be in current working directory

IanMayo:
Fixes #153


rnllv (approved):
Tested on Linux as well

PR #165, opened 2023-04-11
Check for presence of unit_d_star tonal attributes

### LINKED ISSUE #162: Introduce test(s) for extracting tonal data
Our code to extract tonals is now working.

Significantly, it includes steps to allow an element constructed like this to be matched:
```
<strong><span>*</span>Commonly Detected Sources</strong>
```

I have made a test data-file look like this (`france_legacy\unit_d_star.html`).  We should verify that the tonals for this file get successfully processed.

I _guess_ we could also include these tests:
- test for total number of tonals from standard countries
- test for total number of tonals from n-s countries

I'm ok with the above being hard-coded `magic-numbers`.  If/when the test breaks, we consider why it is breaking, and either:
* fix the code to remove what is breaking the test
* fix the text to match a new total, based on new logic

rnllv:
fixes #162

### DIFF
--- a/data/France_Legacy/unit_d_star.html
+++ b/data/France_Legacy/unit_d_star.html
@@ -14,7 +14,7 @@
                 </tr>
                 <tr>
                     <td><strong>Cruise Plant</strong></td>
-                    <td>4 * PR32 Diesel Engines</td>
+                    <td>4 * PR32 Diesel Engines %TEST-162%</td>
                     <td><strong>Shaft x Blade</strong></td>
                     <td>2 * 4 (GRR)</td>
                 </tr>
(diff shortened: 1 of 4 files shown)


IanMayo:
Hello @rnllv - there is a particular test that I want us to perform.

This class has a `*` in the block for `Commonly Observed Tonals`:
image

I want us to check that we still find that table - and that the fancy `find` doesn't break in the future. One way would be to create a specially formatted tonal, and to verify that the tonal has been loaded.

IanMayo (dismissed):

rnllv:
Done.

IanMayo (approved):

PR #167, opened 2023-04-12
166 anchor in country page

### LINKED ISSUE #166: Handle <a> without <href> in ns country list
Our country file contains a list of classes.

Some have links to more data, others don't - they just list the metadata.

One class is an `anchor`, but not a `link`.  So, it has `name` attribute but not `href`.  We should add this to the list of classes for that country.

IanMayo:
Fixes #166

### DIFF
--- a/legacyman_parser/parse_images_of_class.py
+++ b/legacyman_parser/parse_images_of_class.py
@@ -45,7 +45,7 @@ def extract_class_images(soup: BeautifulSoup = None, parsed_url: str = None, par
 
     def not_blacklisted_filter(tag):
         # filter for images that aren't in black-listed images
-        black_list = ['image020.jpg', 'check_db.gif', 'pref_db.gif', 'Rtn2Map_db.gif']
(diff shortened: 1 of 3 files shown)


legacyman_parser/parse_images_of_class.py
```diff
@@ -45,7 +45,7 @@ def extract_class_images(soup: BeautifulSoup = None, parsed_url: str = None, par
 
     def not_blacklisted_filter(tag):
         # filter for images that aren't in black-listed images
-        black_list = ['image020.jpg', 'check_db.gif', 'pref_db.gif', 'Rtn2Map_db.gif']
+        black_list = ['image020.jpg', 'check_db.gif', 'prev_db.gif', 'rtn2map_db.gif']
```
rnllv, line 48: Guess list processing is the reason why our pioneers were adamant on having  a case sensitive file system :grin:

rnllv (approved):

PR #172, opened 2023-04-12
Combine non-standard and standard country classes

### LINKED ISSUE #168: ns countries not shown in DataBrowser
Data for ns countries is not shown in the Data Browser.

I see that the content is pushed to the `publication.js` document, in the `ns_country_classes` element.

I'm pretty sure this element is ignored by DataBrowser.

But, _maybe_ DataBrowser doesn't have to know about `classes` and `ns_country_classes`.  _Maybe_ we just combine them both into one classes array, and that goes into `units`.

rnllv:
fixes #168


rnllv:
Please review

IanMayo (approved):

PR #174, opened 2023-04-13
173 inconsistent exported data

### LINKED ISSUE #173: Inconsistent exported data
I've done a new run-through of the parse, and loaded it into `DataBrowser`.

But, `DataBrowser` has fallen over, and doesn't render anything. Looking at the console, it is reporting that:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'country')
```
I've looked at the code, and it's in the `MockDataProvider`.  In our data provider we inject the missing fields for tonals. First we find the unit, then we fill in supporting fields.  The first one we _expand_ is `correspondingCountry`.

I looked at some of the data.  One unit shows it is from `country: 81`.  But, when I look at the list of countries, it jumps from country `id:80` to country `id:82`.  I _think_ this may be related to us combining the two lists.  _maybe_ we export the list of countries for standard countries, but not those for ns-countries - which could leave gaps.

Yes, there are several countries missing from the `countries` element.  The indication is that there is a _jump_ in the country id that shows the missing _slot_.  These instances are for ns-countries.

Aah, I see the pattern is present in our mock data:
image

Aah, there _may_ be a related issue.  I have a unit that specifies `platform_sub_type_id` as `23`, but the array of `platform_sub_types` only goes up to `22`.  Aah, this is for ns-country too.

rnllv:
fixes #173

### DIFF
--- a/legacyman_parser/parse_countries.py
+++ b/legacyman_parser/parse_countries.py
@@ -24,7 +24,7 @@ def extract_countries_in_region(soup: BeautifulSoup = None,
                                 parsed_url: str = None,
                                 parent_url: str = None,
                                 userland_dict: dict = None) -> []:
-    seq = len(COUNTRY_COLLECTION)
+    seq = userland_dict['seq']
     # get the heading
     heading = soup.find('h1')
     # get the first table after the heading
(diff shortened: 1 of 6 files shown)


rnllv:
Please review.

IanMayo (approved):

IanMayo:
Tested locally, generates valid data

PR #180, opened 2023-04-14
168 convert uniq id assert to diagnostic

### LINKED ISSUE #168: ns countries not shown in DataBrowser

rnllv:
fixes #168


IanMayo (changes requested):
Are there any more diagnostics we could use to help understand the issue.  I _guess_ the name of the country file(s) that caused the duplication could be useful, since one cause of this is one country referring to a class from another country (since both countries purchased the same kit)

rnllv:
>one cause of this is one country referring to a class from another country (since both countries purchased the same kit)

Do you remember if these asserts were failing on real data? If it were failing, then we'll need to replicate the same in mock data. It was failing for Britain in mock data because Britain was being referenced from Europe as well from outside Europe and Britain ended up accumulating two instances of the same class.

The reason we wanted this assert was to give us confidence that we can use a combination of `Country`, `Sub-Category`, and `Unit name` to uniquely identify any class/unit in the entire publication. Once confident, we can use this identification to bookmark these entities in DataBrowser in a persistent manner.

Even if same units are referenced from two or more countries (say in mock data `Unit d` of `French - Legacy` sub-category being referenced from France and Narnia) this assert will not fail as Narnia or France will only have one `Unit d` of `French - Legacy` sub-category.

IanMayo:
yes, I think I did see the assertion on the real data

IanMayo (dismissed):

rnllv:
@IanMayo, commit https://github.com/DeepBlueCLtd/LegacyMan/pull/180/commits/cb7a0f23caf923802e9d586f6afb8c6134195544 will print additional logs like files from where this combination is referenced.

Since the mechanism has to know before hand which all combination has to be watched, this is designed as a two stage process.
First stage, run the parser as is. It'll generate the combinations which are found to have more that 1 occurence.
Second stage, run the parser again after declaring the list `country_category_unit_combination_watcher` in  `legacyman_parser/utils/discrepancy_watcher.py` with the combinations identified in the first stage.

I understand this is convoluted but the other way around would be to run the parser twice automatically.

Please let me know if you need additional info.

IanMayo:
Thanks for the new feature.  But, isn't it an option to capture the extra detail on the single pass, just in case it's required?

rnllv:
Logic modified to produce diagnostics without any previous run input. Output is slightly modified.

rnllv:
Please pull and review.

IanMayo (approved):

IanMayo:
> Do you remember if these asserts were failing on real data?

No @rnllv - we don't see this assert on the real data. I was mistaken.

PR #182, opened 2023-04-16
Create publication.js if not present

### LINKED ISSUE #177: Can't write JSON output if file not already present
This caused some head-scratching.

Our json-publish process only works if there is already a file called `json_publication.js` in the `target` folder.

This is because we open the file in `"r+"` mode:
```
    # Dump the wrapper to the text file passed as argument
    with open(EXPORT_FILE, 'r+') as f:
```

I think we should switch to opening the file in "w" mode.

rnllv:
fixes #177


rnllv:
Please review.

IanMayo (approved):

PR #190, opened 2023-04-22
Capture class data spread across multiple tonal pages

### LINKED ISSUE #150: Class data spread across multiple files
For some classes of platform, there is so much detail known that it is recorded in multiple pages.

So, the propulsion data is on a separate page.  We should reproduce this in our data (done: `unit_split_a.html`).

(Note: I now see other instances where a block of supporting data is in another file.  From what I've seen so far, the structure of the other file matches what we're expecting - so we can use our existing `soup.find()` methods, it's just that we run them on the contents of a different file.

@IanMayo to reproduce the pattern, @AbegaM to implement the handling logic.  Sample data included in #176 

Note: when this happens, sometimes the `colspan=4` table is missing, since it's on the _other_ page.  So, don't fail when this happens, just track it - so we can fix it.

Actually - this is another way of handling `h1_propulsion`  not found in `parse_propulsion`. We currently output an error, instead, do this:
- see if there is a QuickLinksTable
- see if it contains a cell titled `Propulsion`
- get href for `Propulsion`
- check href isn't for current file (either includes current filename, or starts with `#`)
- generate DITA filename for target file (note: you may need to strip the anchor off the end)
  - check if target file already present.  If not:
    - generate DITA target file
    - get content block(s) from source file (**see section below**)
    - convert blocks of content from `htmlToDita`
    - insert converted blocks into DITA target file, as a `reference` object, like [this](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/br_power_units.dita) one. 
- generate link to target (replacing the anchor, if necessary)
- insert href as `propulsionRef` (see [example](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/legacy/unit_legacy_a.dita#L12))

We should also do something like the above for `Remarks`, too.

Note: the demo file for this is [`unit_ab`](https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/data/Britain_Composite/unit_split_a.html) under `Britain_Composites`.  The main body of content is under `unit_split_a.html`, but the `propulsion` and `remarks` sections are missing.  So, first observe that the `colspan=4` table is missing- then get the data from `QuickLinksTable`.

See how the `Propulsion` and `Remarks` links are in different files - and not the current file (which would start with just the anchor (`#remarks`).

## How to find content

See `Finding content section of target file` in #289

rnllv:
fixes #150

### DIFF
--- a/legacyman_parser/utils/class_html_template.py
+++ b/legacyman_parser/utils/class_html_template.py
@@ -1,4 +1,3 @@
-
 class ClassU:
     def __init__(self,
                  id,
@@ -12,7 +11,8 @@ class ClassU:
                  av_temp,
                  reduction_ratio,
                  has_tonal,
-                 tonal_href):
+                 tonal_href,
+                 propulsion_href):
         self.id = id
         self.class_u = class_u
         self.sub_category = sub_category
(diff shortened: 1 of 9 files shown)


legacyman_parser/utils/class_html_template.py
```diff
@@ -12,7 +11,8 @@ def __init__(self,
                  av_temp,
                  reduction_ratio,
                  has_tonal,
-                 tonal_href):
+                 tonal_href,
+                 propulsion_href):
```
IanMayo: Hello @rnllv - we only call `ClassU` from one place, and in that place we pass a value of `None` for `propulsion_href`.  So, howabout we remove it from the constructor, and set the initial value of `self.propulsion_href` to `None`.

rnllv: Okay

rnllv:
Please review.

IanMayo (approved):

PR #191, opened 2023-04-23
Parse class attributes from tonals page in a colspan insensitive way

### LINKED ISSUE #178: Class attributes from colspan-4 table
I've encountered at least 6 forms of the colspan=4 class attributes table.

We currently handle this table through the `parse_class_attributes_from_tonals` module.  But, that relies on a specific table arrangement. Sadly the real tables aren't that consistent.

The tables have varying combinations of merged cells.  Sometimes there are rowspans of 2, 3 or 4.  But, the cells always appear to be in `label`, `value` pairs.

So, I _think_ I need to collate a lookup table. We work our way through the pairs, and lookup the field to use to store the labelled item.  When the lookup table doesn't contain the label, throw a warning and I'll steadily implement them.

Here is an example:
image

So the lookup would be:
* Cruise Plant: CruisePlant
* Boost Plant: BoostPlant
* Reduction Ratio: ReductionRatio
* R:R: ReductionRatio
* Max TRPM: MaxTRPM
* TRPM Range:  MinTRPM, MaxTRPM
* Shaft x Blade: ShaftBladeProps
* MTBF: MTBF
* BHP: BHP
* Max SRPM: MaxSRPM
* SRPM Range: Min SRPM, Max SRPM

Ian should create some variety in the mock data, to _drive_ this.

rnllv:
fixes #178


rnllv:
Please review.

IanMayo (approved):

PR #192, opened 2023-04-23
Cache html content on first read

### LINKED ISSUE #186: Option to reduce data volume
As we process the real data, the parse takes a few minutes (around 2).

Once the processing is complete the compiled JSON brings VS-Code performance challenges. In particular it can take some time before it's possible to collapse JSON tree elements.

When I'm doing compile/test/inspect cycle this becomes time consuming.

One solution could be to only process a fixed number of countries (or classes).  This wouldn't work with our mock data, since we have such a small set of data. But, we could process a proportion of the data (25%).

rnllv:
Supports #186

### DIFF
--- a/crawler/simple_crawler.py
+++ b/crawler/simple_crawler.py
@@ -5,6 +5,8 @@ from urllib.parse import urlparse, urljoin
 
 from bs4 import BeautifulSoup
 
+from legacyman_parser.utils.html_cacher import check_if_html_cached, get_html_cache, set_and_return_html_cache
+
 
 class Response:
     def __init__(self, status_code, reason):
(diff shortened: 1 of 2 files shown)


crawler/simple_crawler.py
```diff
@@ -14,11 +16,17 @@ def __init__(self, status_code, reason):
...
+        response = Response(200, 'OK')
+        response.text = get_html_cache(url)
+    else:
+        status_code, reason = check_path(url)
+        response = Response(status_code, reason)
     if response.status_code == 200:
-        with open(url, 'rb') as f:
-            response.text = f.read()
+        if response.text is None:
+            with open(url, 'rb') as f:
```
IanMayo: doesn´t `rb` mean read in binary mode?

Aah, while HTML files are text, images are binary.  But, I _think_ we only ever copy binary files - I don´t think we open them.

rnllv: Yep, let me change it to 'r'

rnllv:
Please review.

IanMayo (approved):

PR #201, opened 2023-04-27
196 multiple quicklinkstable in class file

### LINKED ISSUE #196: Multiple QuickLinksTable in class file
For some class files with a lot of detail, they are broken down into several sections - each with its own `QuickLinksTable`

Each table contains links other parts of the document (sometimes in another HMTL document).  But, the table cell for _this_ section isn't clickable - it's plain text. 

When we are looking for the destination of the `Propulsion` data, we check the URL for the `Propulsion` entry.  If we are looking for the `Propulsion` URL we need to cycle through `QuickLinksTable`, `QuickLinksTable2`, `QuickLinksTable3`, `QuickLinksTable4`

We could do this either by a list of strings, or providing a wildcard.  Hmm, I _guess_ we would `find_all` for the wildcard version of the id, then do a `filter` to find the first one with a `td` of `propulsion` that is in a `<a>` element.

**Update** suitable data for this was added in #198

rnllv:
fixes #196

### DIFF
--- a/data/Britain_Composite/unit_split_a.html
+++ b/data/Britain_Composite/unit_split_a.html
@@ -21,6 +21,12 @@
       </tr>
     </table>
   </div>
+  <div id="QuickLinksTable1">
+  </div>
+  <div id="QuickLinksTable2">
+  </div>
+  <div id="QuickLinksTable3">
+  </div>
 </div>
 <div id="BottomLayer">
   <div id="PageLayer">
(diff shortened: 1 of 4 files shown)


IanMayo:
Note: I setup [this](https://github.com/DeepBlueCLtd/LegacyMan/pull/198/files#diff-57b90e7aa3b4982aefe06f72c6dbfd8f84b496bfef16a5616bba85180b2edf33L2) data file with multiple propulsion links, matching the pattern in the real data.

But, I didn't include the changing div ids.  I'll modify that file in this PR, so we further test the logic.

IanMayo (approved):

PR #205, opened 2023-04-30
String unit ids

### LINKED ISSUE #171: Change id of units/classes to string instead of number sequence
Currently, id for country/sub-category/class/unit is a sequence number. This has to be changed to a reproducible string id so that URLs (and potentially favourites saved as cookies) still work. This will also support https://github.com/DeepBlueCLtd/DataBrowser/issues/22 as the assert in legacyman_parser/parser.py line 268 ensures that class name, sub-category, and country name combinations are unique.

I'm pretty sure we need to test our assumption of unique ids in the parse process, and to come up with a pattern of how to shorten the term.  So, maybe we need to test unique shortened ids, as well as unique ids.

rnllv:
fixes #171
Also, fixes https://github.com/DeepBlueCLtd/DataBrowser/issues/22

### DIFF
--- a/legacy_tester/test_payload/mock_data_test_payload.json
+++ b/legacy_tester/test_payload/mock_data_test_payload.json
@@ -23,7 +23,10 @@
                                 "expected_count": 2
                               }],
   "check_presence_of_common_class_images_in_different_class_folder": [{
-                                                                        "unit_ids": [8, 4],
+                                                                        "unit_ids": [
+                                                                                      "*Unit_dLegacyNarnia",
+                                                                                      "Unit_dLegacy - TraditionalFrance"
+                                                                                    ],
                                                                         "image_name_contains": "NEW-TEST-48"
                                                                       }],
   "check_presence_of_classes_of_non_standard_countries": {"name": "%TEST-106 Icelandic Comp", "count": 2},
--- a/legacyman_parser/parser.py
+++ b/legacyman_parser/parser.py
@@ -124,7 +124,7 @@ def parse_from_root():
 
     watched_unit_category_country_combination_discrepancy_collector = {}
     print("\n\nParsing Classes:")
-    standard_class_parser = ClassParser(0, {})
+    standard_class_parser = ClassParser({})
     for country in COUNTRY_COLLECTION:
         """Parsing classes in each country"""
         if country.url is None:
@@ -152,8 +152,7 @@ def parse_from_root():
                                                                            len(COUNTRY_COLLECTION)))
 
     print("\n\nParsing Classes from non-standard countries:")
-    ns_class_parser = ClassParser(len(
-        standard_class_parser.CLASS_COLLECTION), standard_class_parser.SUBTYPE_COLLECTION)
+    ns_class_parser = ClassParser(standard_class_parser.SUBTYPE_COLLECTION)
     for ns_country in NON_STANDARD_COUNTRY_COLLECTION:
         if ns_country.url is None:
             INVALID_COUNTRY_HREFS.append({"country": ns_country.country,
--- a/legacyman_parser/utils/parse_class_table.py
+++ b/legacyman_parser/utils/parse_class_table.py
@@ -1,3 +1,4 @@
+from random import random
 from urllib.parse import urljoin
 
 from bs4 import BeautifulSoup, PageElement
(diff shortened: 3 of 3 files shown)


rnllv:
This might probably be not required as https://github.com/DeepBlueCLtd/DataBrowser/pull/75 will supersede this if approved.

IanMayo:
> This might probably be not required as [DeepBlueCLtd/DataBrowser#75](https://github.com/DeepBlueCLtd/DataBrowser/pull/75) will supersede this if approved.

No - I think we should still have human-readable IDs, particularly if the ID is the filename.

rnllv:
Do you see any use for such string ids, @IanMayo, given that bookmarks are human-readable as of now? Might be like user 1 informing user 2 "on-the-field" that _<string-hash-id/human-readable-id> is approaching us_.

I'm only seeing  increased storage costs.

IanMayo:
I don't think there is an end-user benefit in having human-readable class ids.

But, there are non-end-user benefits:
1. The DITA maintainer will be working in a file-based editing environment. For him, he needs human readable filenames to know which one to open.  Aah, when linking to other classes/countries/categories the author will need to be able to read/understand the filenames in order to navigate to the correct one.
2. While debugging the exported `publication.js`, having human-readable class (or other) ids would make things easier to navigate.   This isn't necessarily a short-term development-related issue.  It _could_ be a factor in the long term maintenance of FieldMan.


_Hopefully_ we can analyse the data and identify the minimum number of characters we need to uniquely represent countries, categories, and classes - which will keep the identifiers short.   Each time we process the DITA export we can double-check the above.

Or, can you think of/suggest any other strategies?

rnllv:
>The DITA maintainer will be working in a file-based editing environment. For him, he needs human readable filenames to know which one to open. Aah, when linking to other classes/countries/categories the author will need to be able to read/understand the filenames in order to navigate to the correct one.

Got it. But I suggest let this be a new field. Like 'class_ref_id' or 'cntry_ref_id' etc, but not be used for relating records from child tables/groups. Let the existing id do that.

IanMayo:
> Got it. But I suggest let this be a new field. Like 'class_ref_id' or 'cntry_ref_id' etc, but not be used for relating records from child tables/groups. Let the existing id do that.

That meets benefit 1, but not benefit 2.  I accept that the wordy id could be `fragile` - as it's sensitive to the metadata changing.  But if we are going to have a persistent id value, do you have a plan of how to produce it?  If it's a combination of these metadata fields, it's just as fragile.

We're trying to minimise the file-size.  I've raised a worry about the challenge in navigating a large JSON file in VS-Code. But, I think that's more about the number of lines and number of objects, than the length of strings.  Well, I _think_ it is...

rnllv:
>That meets benefit 1, but not benefit 2. 

Nope. Let me be more descriptive.
```
class: 
 Name: Billowing Water
 id: 94 [Generated sequentially, essentially primary key in a relational sense]
 ref_id: "Britain:WestShip:GestShip:BillowWat" [yet to finalize the string shortening logic]
 other_attr: ...
 ...
 ...
 ...
tonal: 
 Name: SonarSense
 id: 153 [Generated sequentially, essentially primary key in a relational sense]
 class_id: 94 [Foreign key reference]
 ref_id: "Britain:WestShip:GestShip:BillowWat:SonarSense1" [yet to finalize the string shortening logic]
 other_attr: ...
 ...
 ...
 ...
```

We can have these descriptive ids for tonals as well (at the cost of json size). What I'm saying is let joins be dependent on the number sequence that we generate.

> I accept that the wordy id could be fragile - as it's sensitive to the metadata changing. But if we are going to have a persistent id value, do you have a plan of how to produce it? If it's a combination of these metadata fields, it's just as fragile.

I believe we're supposed to factor in this fragility into the id generation logic because there's nothing we know about the data that we can consider to be permanent/fixed. An id generation logic that's dependent on the (fragile) values of the metadata will be helpful for DITA maintainers as well as they'll get to know instantly what id to look for if they change metadata. They'll also know why a bookmark is not working if they care to inspect the localStorage from their browsers.

rnllv:
@IanMayo, shall I go ahead and add `ref_id`s for classes and tonals? That will help us visualize these changes.

IanMayo:
> @IanMayo, shall I go ahead and add `ref_id`s for classes and tonals? That will help us visualize these changes.

yes please :-)

rnllv:
I'll close this PR and create a new branch for `ref_id` changes.

PR #215, opened 2023-05-06
Handle 7 columns tonal table

### LINKED ISSUE #212: 7 column tonal table
Oh great.  I have just encountered a 7 column tonal table.

We should reproduce it in our data, and check we handle it.

There is now data in `develop` that reproduces the layout. We should update our parser to read it, and include that in a test.

I've just looked at our code, and _maybe_ it's not important. We just need to test that.

rnllv:
fixes #212


rnllv:
Please review.

IanMayo (approved):

PR #219, opened 2023-05-11
171 create reference id for classes and tonals

### LINKED ISSUE #171: Change id of units/classes to string instead of number sequence

rnllv:
fixes #171 

This is a long form id. Once we're ready with our switchable maps, we should be able to condense the components in the id string.

### DIFF
--- a/legacyman_parser/utils/ref_id_generator.py
+++ b/legacyman_parser/utils/ref_id_generator.py
@@ -15,7 +15,7 @@ def get_ref_id_for_class(classu):
     # TODO: Introduce string shortening logic
     return (classu.country.country + ":"
             + classu.sub_category[0] + ":"
-            + classu.class_u).lower().replace("-", "").replace(":", "-")
+            + classu.class_u).lower().replace(" ", "_").replace(":", "-")
 
 
 def get_audited_ref_id_for_class(classu):
--- a/legacyman_parser/utils/ref_id_generator.py
+++ b/legacyman_parser/utils/ref_id_generator.py
@@ -1,6 +1,5 @@
 import re
 
-tonal_map = {}
 # Name to start with alphabet, and then should contain one
 # or more of the following characters
 # a to z
@@ -10,7 +9,6 @@ tonal_map = {}
 accepted_ref_id_pattern = re.compile(r"^[a-z][a-z0-9-_]+$")
 
 classes_with_illegal_reference_id_characters = []
-tonals_with_illegal_reference_id_characters = []
 
 
 def get_ref_id_for_class(classu):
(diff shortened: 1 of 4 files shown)


IanMayo (commented):
is it worth using a separator that is legitimate in a filename, such as "-"

Then we have the _option_ of using it as a filename.

rnllv:
>is it worth using a separator that is legitimate in a filename, such as "-"


Depends on how this id is being used.

If this is just for reference, then there shouldn't be an issue. We can use the same as filenames as well.
But if we're required to parse this reference id to its component country, category, and sub-category, then we might have a problem (based on how we design our id string contraction logic ) as few category/sub-category already contain a "-" in them. 

That's the only problem I see now.

rnllv:
@IanMayo, please check if this suffices https://github.com/DeepBlueCLtd/LegacyMan/pull/219/commits/0afeaa7c7e3987f37887864d7f39505836a3e75a

IanMayo:
I may be wrong on this, but I don't think we need unique, human-legible, persistent IDs for tonals.    It's the classes that get persisted as a favourite, and it's the classes that we expect to be the `unit of storage` (file).

There is common duplication in the tonals table. I think I've seen `Unk` for 5 consecutive rows.  This will prevent the ids being unique (we'll need a counter) or persistent (inserting a row would change the id).

Sorry I didn't notice this from the issue/PR title :-(. I only realised it when I was looking at the updated `publication.js`.

rnllv:
No issues, @IanMayo. I'll revert the changes for tonals.

rnllv:
@IanMayo, please check now.

legacyman_parser/utils/ref_id_generator.py
```diff
@@ -0,0 +1,25 @@
...
+accepted_ref_id_pattern = re.compile(r"^[a-z][a-z0-9-_]+$")
+
+classes_with_illegal_reference_id_characters = []
+
+
+def get_ref_id_for_class(classu):
+    # TODO: Introduce string shortening logic
+    return (classu.country.country + ":"
+            + classu.sub_category[0] + ":"
+            + classu.class_u).lower().replace(" ", "").replace("-", "").replace(":", "-")
```
IanMayo: for our first version, could we not remove the spaces? Since we're converting it all to lower case, some word combinations may not be human-guessable.

rnllv: We could, but spaces in filenames are not handled gracefully by a lot of tools. Please check https://github.com/DeepBlueCLtd/LegacyMan/pull/219/commits/05390c10dc27f1f5ec1387fba8d594785627ccb6

IanMayo: How about keep hyphens, and swap spaces for underscores?

rnllv: Okay, retain hyphens, use hyphens as separator for different components, and replace spaces with underscore. Please check https://github.com/DeepBlueCLtd/LegacyMan/pull/219/commits/1023c24a1c89c958c7aa2af665856aaa432cfba8

IanMayo (approved):

PR #243, opened 2023-07-03
240 Parse more data from regions, for new world map page in DITA

### LINKED ISSUE #240: First step(s) in export to DITA - parse world map (regions)
We will start off by generating the world map page.

The current python code finds the world-map page, and passes it to `extract_regions`.

We will extend the `extract_regions` code to pull in this extra data which we will require:
- the URL of the world-map image
- the coordinates for each URL on the world map

So:
* [ ] modify `parser.py` to exit python after we have processed the regions (to avoid having to wait for all of the other processing to run)
* [ ] modify `RegionMap` in `extract_regions` so it can store a series of sets coordinates
* [ ] modify `extract_regions` so it has a local variable to store the world-map image URL
* [ ] modify `extract_regions` to extract and then store the URL of the world-map image
* [ ] modify extract_regions` to store the coords of each region
* [ ] modify `parser.py` to output the URL plus full set of regions once the map file is parsed.

No new files need to be generated at this stage.  We're just parsing content, and showing it on the console.

brahnavank:
Fixes #240 
Changes to add extra data to regions

### DIFF
--- a/legacyman_parser/parse_non_standard_countries.py
+++ b/legacyman_parser/parse_non_standard_countries.py
@@ -10,7 +10,6 @@ to process non-standard country data
 """
 NON_STANDARD_COUNTRY_COLLECTION = []
 
-
 def extract_non_standard_countries_in_region(soup: BeautifulSoup = None,
                                              parsed_url: str = None,
                                              parent_url: str = None,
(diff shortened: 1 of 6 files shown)


legacyman_parser/parse_non_standard_countries.py
```diff
@@ -19,8 +18,10 @@ def extract_non_standard_countries_in_region(soup: BeautifulSoup = None,
     for area_element in soup.find_all('area'):
         href = area_element.get('href')
         country = area_element.get('alt')
+        coords = area_element.get('coords')
+        shape = area_element.get('shape')
         if href.startswith('..'):
             parsed_region_url = urljoin(parsed_url, href)
             NON_STANDARD_COUNTRY_COLLECTION.append(CountryMap(seq.next_value(), country,
```
IanMayo, line 25: @brahnavank - while the next block of code refers to `RegionMap` I don´t think it´s actually a  region object.  I think that in this instance a plain old URL is being stored in region-map.

But, now that `RegionMap` includes type and cooridinates we should probably change the code below to use a  new data object (that looks like the original RegionMap structure).

I can have a go at this tomorrow.

IanMayo:
There is also this requested change, please:
image

legacyman_parser/parse_non_standard_countries.py
IanMayo: No, I was wrong. We are parsing the map twice - once for standard countries, and once for non-standard countries.

PR #248, opened 2023-07-04
Changes to create regions.dita

### LINKED ISSUE #241: Export world map
The point of this issue is to generate a DITA regions file, like this one:
https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/regions.dita

* [ ] introduce `publish regions` method - it will receive a map URL, a list of regions, and a `target` folder URL
* [ ] if not already present, create the `target` folder
* [ ] copy the map image file from the URL to a `content\images` folder under `target/dita`.
* [ ] create a `regions.dita` file in the `target\dita` folder
* [ ] fill in regions the data for `regions.dita`

brahnavank:
Fix #241 
Changes to create regions.dita


legacyman_parser/parser.py
```diff
@@ -90,7 +89,9 @@ def parse_from_root():
     for r in NON_STANDARD_COUNTRY_COLLECTION:
         combined_regions.append(r.region)
 
-    publish_regions(REGION_DATA.url, combined_regions, "dita_out")
+    # publish_regions(REGION_DATA.url, combined_regions, "dita_out")
+    published_dita = dita_publisher.publish_regions(parsed_regions=REGION_DATA)
```
IanMayo: we don need to store the results of publish_regions

legacyman_parser/dita/create_regions.py
```diff
@@ -0,0 +1,62 @@
+from xml.dom import minidom
```
IanMayo: could you move this code into `legacy_publisher\dita_publisher`, since itś related to publishing dita data, rather than parsing data

IanMayo:
could you double-check the to-do items in the issue please?

legacy_publisher/dita_publisher.py
```diff
@@ -1,5 +1,71 @@
...
-        print("outputting region", reg.shape, reg.url, reg.coords)
\ No newline at end of file
+from xml.dom import minidom
+from os.path import dirname, abspath
+from legacyman_parser.utils.constants import DITA_REGIONS_EXPORT_FILE
+
+"""This module will handle post parsing enhancements for DITA publishing"""
+DITA_EXPORT_FILE = DITA_REGIONS_EXPORT_FILE
+
+random.seed(100)
```
IanMayo: @brahnavank  - we don´t need or use random numbers, I guess this is a copy/paste error.  Please copy ideas, but not code.

legacy_publisher/dita_publisher.py
```diff
@@ -1,5 +1,71 @@
...
+import os 
 
-def publish_regions(map_url, regions, target_folder):
-    print('outputting regions', map_url, target_folder)
-    for reg in regions:
-        print("outputting region", reg.shape, reg.url, reg.coords)
\ No newline at end of file
+from xml.dom import minidom
+from os.path import dirname, abspath
+from legacyman_parser.utils.constants import DITA_REGIONS_EXPORT_FILE
```
IanMayo, line 6: there is just one file called `regions.dita`  It will always be called `regions.dita`.  We can hard-code the issue.

target/data/regions.dita
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
```
IanMayo: we don´t need this file - just the one in `dita`.

legacy_publisher/dita_publisher.py
```diff
@@ -64,5 +66,7 @@ def publish_regions(regions=None):
     if not isExist:
         os.makedirs(dita_dir)
 
+    copy_tree(sourcepath+"/PlatformData/content", dita_dir+"/content")
```
IanMayo: let´s not copy over the content directory.  It may not be called content., and we don´t necessarily need all of the files.  Let´s get the URL of the image from `regions.url`, and copy the image to the destination.

target/dita/regions.dita
```diff
@@ -0,0 +1,30 @@
...
+<?xml version="1.0" ?>
+<topic id="links_1">
+	<title>Regions</title>
+	<body>
+		<imagemap>
+			<image href="./Content/Images/world-map.gif"/>
+			<area>
+				<coords>540,326,611,308,681,326,711,286,659,128,618,139,588,199,572,273</coords>
+				<shape>poly</shape>
+				<xref href="../../../target/dita/regions.dita" format="">Europe</xref>
```
IanMayo: Aah, @brahnavank - the URLs won't have valid destinations, but we should still include the existing HTML references (like `xxxx/europe.html`) as a prompt for what is going to go in there

IanMayo:
@brahnavank - Oxygen is complaining that the XML processing instruction is in the wrong place.  It comes before the DOCTYPE in our demo file for `regions.dita`.
image

legacy_publisher/dita_publisher.py
```diff
@@ -66,7 +69,17 @@ def publish_regions(regions=None, sourcepath=None):
     if not isExist:
         os.makedirs(dita_dir)
 
-    copy_tree(sourcepath+"/PlatformData/content", dita_dir+"/content")
+    # copy map image
+    image_url = abspath(regions.url);
+    target_dir = dirname(dirname(dirname(abspath(DITA_REGIONS_EXPORT_FILE))))
```
IanMayo, line 74: can you teach me why we are calling `dirname` 3 times?

IanMayo:
The exported content is rejected by oxygen:
image

IanMayo (approved):

PR #249, opened 2023-07-05
Changes to add validation

### LINKED ISSUE #242: Validate world map in DITA
Now that we are generating the world map, extend `publish regions` to validate the XML in the generated map.

It's ok to introduce a `dtd` at the root of the `LegacyMan` repo to contain a copy of the schemas.

**Update:**  @brahnavank has established that we can't use LXML for validation, since it can't traverse the complex schema constructs we use for DITA (and our DITA specialisation).

So, we'll use `dita-ot` for validation.  But, I don't want to incorporate DITA-OT into our repo (yet).  So, we'll install DITA-OT separately on the machine, and ensure `dita` executable is in the path for the terminal.  Actually - it will probably be best to include the DITA-OT installed within Oxygen on the command-line `path`, since that's the version that will be doing the `publish`.

Plus, we'll find a way of ensuring that the `publish` process fails/exits when it encounters invalid XML/dita.

brahnavank:
Fixes #242
Changes to add validation


IanMayo (commented):
ok. the script won't work for my machine, since I won't have dita-ot in that path, as required for the executable and the DTD. 

So, could you make our script read a DITA-OT environment variable? Oh it should throw an error if the end variable is unset.

legacy_validator/dita_ot_validator.py
```diff
@@ -0,0 +1,36 @@
+import subprocess
+import os
+
+# DITA-OT global path
+dita_ot_path = '/Users/workspace_blue_python/update_3/LegacyMan/legacy_validator/dita-ot-3.7.2/bin/dita'
```
IanMayo: we can delete this line with a local path, can't we?

legacy_validator/dita_ot_validator.py
```diff
@@ -0,0 +1,36 @@
...
+        print("Command executed successfully.")
+    else:
+        error_message = completed_process.stderr
+        print("Error occurred during command execution:")
+        print("DITA-OT ERROR : ", error_message)
+    
+
+# Path to the XML and DTD files
+xml_file = './dita/regions-invalid.dita '
+dtd_file = './dita-ot-3.7.2/plugins/org.oasis-open.dita.v1_2/dtd/technicalContent/dtd/topic.dtd'
```
IanMayo: we should use env variable for this path.

legacy_validator/dita_ot_validator.py
```diff
@@ -0,0 +1,36 @@
+import subprocess
+import os
+
+# DITA-OT global path
+dita_ot_path = '/Users/workspace_blue_python/update_3/LegacyMan/legacy_validator/dita-ot-3.7.2/bin/dita'
+
+#os.environ["DITA-OT"] = dita_ot_path
+dita_ot_path = os.environ['DITA-OT']
+
```
IanMayo: if we are going to use DITA-OT env variable for both the executable and the schemas, the variable needs to go to the DITA-OT folder, not the executable.  So, when we generate `command` we need to append `bin\dita` to the path, and some other folders for the `dtd_file` (below).

legacy_validator/dita_ot_validator.py
```diff
@@ -0,0 +1,36 @@
...
+# function to validate dita files
```
IanMayo: Another env variable request.  If the variable is unset, could the app exit with a related error message - otherwise it could fail for unexplained reasons. Message something like: `Exiting, since parser relies on DITA-OT environment variable pointing to DITA-OT folder`.

IanMayo: And another env variable request. Could you insert this comment?
```
# the DITA-OT env variable should point to a folder like this:
# /Applications/Oxygen XML Editor/frameworks/dita/DITA-OT3.x
```

IanMayo:
Oh, there's another missing bit.  We need to include the validation in the parsing process.

So, I suggest the `dita_ot_validator' module is moved into `legacy_publisher`, and we validate the file once we have written regions to the hard disk.

legacyman_parser/parser.py
```diff
@@ -91,6 +91,9 @@ def parse_from_root():
 
     dita_publisher.publish_regions(regions=REGION_DATA, sourcepath=cleansed_url+"/PlatformData/PD_1.html")
 
+    #validate
```
IanMayo: we should be calling `validate_regions` directly please - so we pass the filename and the DTD to validate against.

IanMayo:
Remaining items:
* [x] comment out local path constant
* [x] throw error if DITA_OT environment variable isn't set
* [x] include comment with example of DITA-OT path from oxygen
* [x] modify code to accept DITA_OT path that includes spaces
* [x] modify code to pass file to test and DTD to the function
* [x] move the validation code into `legacyman_parser` package
* [x] copy `regions_invalid.dita` into `legacyman_parser` package (for testing)
* [x] remove `legacy_validator` package

IanMayo (dismissed):

IanMayo (approved):

PR #251, opened 2023-07-06
244. Changes to create regions dita

### LINKED ISSUE #244: Create standard regions and countries
* [ ] move the `exit` call down until after countries have been parsed
* [ ] create new `publish_region` method
* [ ] loop through regions
* [ ] create folder for region
* [ ] create a topic file in each region
* [ ] include table of links to child countries in each topic file
* [ ] authenticate as dita `topic`

brahnavank:
Fixes #244 
Changes to create regions dita


legacy_publisher/dita_publisher.py
```diff
@@ -79,14 +77,46 @@ def publish_regions(regions=None, sourcepath=None):
     isDestExist = os.path.exists(dirname(dita_dir+new_path))
     if not isDestExist:
         os.makedirs(dirname(dita_dir+new_path))
-    copy_command = 'cp '+source_path+new_path+' '+dita_dir+new_path
-    os.system(copy_command)
+    os.system('cp '+source_path+new_path+' '+dita_dir+new_path)
 
     with open(DITA_REGIONS_EXPORT_FILE, "w") as f:
        f.write(xml_string_with_doctype)
```
IanMayo, line 86: in the next block, the validation of the regions code has been removed. We need it.

legacy_publisher/dita_publisher.py
```diff
@@ -79,14 +77,46 @@ def publish_regions(regions=None, sourcepath=None):
...
+                entry.appendChild(xref) 
+                row.appendChild(entry)
+
+        isDestExist = os.path.exists(dirname(export_dita))
+        if not isDestExist:
+            os.makedirs(dirname(export_dita))
+        print("export_dita :", export_dita)
+        write_dita_doc(root, export_dita)
+
+
```
IanMayo, line 132: could we also validate the exported dita files please?

IanMayo:
There's a strange behaviour that I don't really recognise. 

The publish process should produce a `target` folder. So, I deleted the `target` folder. Then the publish process fell over, because it couldn't find some flag images.

I _think_ we need to pull forward a block of code in the top level of `parser.py` - code that collates a folder of flags in the target directory.

If you're able to spot this, it would be useful.  Otherwise I'll have a look tomorrow.

legacy_publisher/dita_publisher.py
```diff
@@ -124,6 +124,11 @@ def publish_country_regions(regions=None, stcountries=None, nstcountries=None):
         print("export_dita :", export_dita)
         write_dita_doc(root, export_dita)
 
+        if(dita_ot != None):
+            xml_file = abspath(export_dita)
+            dtd_file = dita_ot+'/plugins/org.oasis-open.dita.v1_2/dtd/technicalContent/dtd/topic.dtd'
```
IanMayo, line 129: we should make the path to the topic.dtd a constant. We're using it in two places so far, we're bound to use it in more places too.

legacyman_parser/parse_flag_of_country.py
```diff
@@ -115,7 +115,8 @@ def extract_flag_of_ns_country(soup: BeautifulSoup = None, parsed_url: str = Non
     # Copy to target directory
     destination_file = COPY_FLAGS_TO_DIRECTORY + \
         userland_dict['country'].country.lower() + os.path.splitext(flag_image)[1].lower()
-    shutil.copy2(flag_image, destination_file)
+    # shutil.copy2(flag_image, destination_file)
```
IanMayo, line 118: hello @brahnavank - is there a reason why we're using the operating system copy instead of the Python one?

Hmm, actually... since the target operating system is MS-Windows we'll need to use the Python version.

I think that the `dita` validate call will still work though, since I'm pretty sure it will just pick up the `dita.bat` version.

IanMayo (approved):

PR #255, opened 2023-07-11
Changes to add collections

### LINKED ISSUE #246: Create collections for non-standard country
Our non-standard country (Britain) has a table of image links.

Process the file as follows
* [ ] introduce a publish_ns_country method, this will receive the country name, the list of links and labels, and a target folder
* [ ] create one sub-folder for each link title (to include country-name and category, `britain-legacy`).
* [ ] introduce a publish_collection method, this will receive a category name the URL of the collection file and the target folder name
* [ ] this method will open the URL, and copy the `classlist` table to the destination (converting to CALS)

brahnavank:
Fixes #246 

Changes to add collection for non standard countries.


legacyman_parser/parse_countries.py
```diff
@@ -98,12 +110,39 @@ def extract_nst_countries_in_region(soup: BeautifulSoup = None,
...
     COUNTRY_TABLE_FOUND.append(RichCollection(title.text, body, None, url, cols, rows))
 
+def extract_collections_non_standard_country(soup: BeautifulSoup = None,
+                                parsed_url: str = None,
+                                parent_url: str = None,
+                                userland_dict: dict = None) -> []:
+
+    url = userland_dict['url']
+    title = soup.find('h2')
+    div_element = soup.find('div', id='PageLayer')
```
IanMayo: sorry, we can't just look for `PageLayer`.  That can appear multiple times on the page.

We have to look for a `td` with a `colspan` of `7`. Then find that element's parent table.  That's the table we need to process.

target/dita/regions/Britain_Complex/Britain_Complex.dita
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE classlist SYSTEM "../../../../dtd/classlist.dtd">
+<classlist id="britain-complex">
+  <title>Britain - Complex</title>
+  <flag>
+    <fig>
+      <image href="#" alt="flag"/>
```
IanMayo: we're still waiting for the above flag URL

target/dita/regions/Britain_Composite/Britain_Composites.dita
```diff
@@ -0,0 +1,55 @@
...
+            <entry>Class</entry>
+            <entry>Designator</entry>
+            <entry>Propulsion</entry>
+            <entry>Shaft x Blade</entry>
+            <entry>Max Speed</entry>
+            <entry>Av TPK</entry>
+            <entry>Reduction Ratio</entry>
+          </row>
+          <row>
+            <entry>Unit_ab</entry>
```
IanMayo: unit_ab should have a URL, like in the source data:
https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/data/Britain_Composite/Britain_Composites.html#L29

legacyman_parser/parse_countries.py
```diff
@@ -98,12 +123,45 @@ def extract_nst_countries_in_region(soup: BeautifulSoup = None,
...
+                                parent_url: str = None,
+                                userland_dict: dict = None) -> []:
+
+    url = userland_dict['url']
+    flag = userland_dict['flag']
+    flag_dest = userland_dict['flag_dest']
+    
+    title = soup.find('h2')   
+    body = soup.find(lambda tag: tag.name=='table' and tag.find_all('tr')[0].find_all('td')[0].get('colspan') == '7')
+    if body is None:
```
IanMayo: I think there is a simpler version of the above `find` call, which may fail if the wide column isn't in the first row..
```
wide_col = soup.find_all("td", {"colspan" : "7"})
if wide_col is None:
    # do exit
body = wide_col.find_parent("table")
```

target/dita/regions/Britain_Complex/Britain_Complex.dita
IanMayo: Fixed.

target/dita/regions/Britain_Composite/Britain_Composites.dita
IanMayo: Unit_ab should have a URL to the next level down:
image

target/dita/regions/Britain_Complex/Britain_Complex.dita
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
IanMayo: Files at this level should be in `regions/Britain/Complex`.

So all these category folders are inside `Britain`.

At the top level of regions we will just have `Britain`, `Iceland`, `Europe`, `Africa`.

Here is our target folder layout:
image

IanMayo (changes requested):

IanMayo:
You over-rode my fixes here:
image

IanMayo (approved):

PR #268, opened 2023-07-24
Add function to generate dita image-map from html image-map

### LINKED ISSUE #263: Process Regions
Create a DITA imagemap to match the one in `PD_1.html`.  It should use the same image, and the same coordinates.

The source image map will always be in  `.\PlatformData\PD_1.html` under the data folder specified in the attributes.  The generated file will be in `target\dita\regions\regions.dita`.   

The DITA code will match that provided [here](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/regions.dita)

The background image of the world will be copied to `target/dita/regions/content/images/world-map.gif`

The region/country name will come from the `Alt` attribute, and the target hrefs will be constructed as `./regions/regions.dita` (that's the current file, since we haven't generated the target files yet).  The actual target files will created in a future step (#264).

ebenezer-melkamu:
Fixes #263 

### Modifications
1. Added a function to copy the `world-map.gif` file from the `data` folder to `regions` folder


ebenezer-melkamu:
@IanMayo Should i use libraries like `ElementTree` to create the `PD_1.html` file from my python code, something like this

```py
  
import xml.etree.ElementTree as ET

# create the root element
root = ET.Element("map")
root.set("name", "image-map")

# create the image element
img = ET.Element("image")
img.set("src", "./Content/Images/world-map.gif")
img.set("width", "800")
img.set("height", "600")
root.append(img)

```

IanMayo:
> @IanMayo Should i use libraries like `ElementTree` to create the `PD_1.html` file from my python code, something like this 

yes, but you are creating `world-map.dita`.  The `.dita` files have a slightly different format for some elements, so check out the mock version I've linked to in the issue.

ebenezer-melkamu:
@IanMayo, I have successfully generated the `regions.dita` file from the `html` file. Can you please provide me with some guidance for the final step? Do I need to create folders in the `regions` directory based on the `alt` attribute of the country names?

target/dita/regions/regions.dita
```diff
@@ -0,0 +1,48 @@
...
+  <imagemap>
+   <image href="./Content/Images/world-map.gif"/>
+   <area>
+    <shape>
+     poly
+    </shape>
+    <coords>
+     550,211,522,231,515,263,540,275,564,269,561,231,555,214
+    </coords>
+    <xref format="dita" href="../Britain/Britain1.html"/>
```
IanMayo: this XREF will fail, since there isn't an html file at `../Britain/Britain1.html`.  Instead, we'll just link to the current file, so it's `./regions.dita`.

We'll create these country folders (and files) in the next issue.

target/dita/regions/regions.dita
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+&lt;!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd"&gt;
+<topic id="links_1">
+ <title>
+  Regions
+ </title>
+ <body>
+  <imagemap>
+   <image href="./Content/Images/world-map.gif"/>
```
IanMayo: the parser needs to include the step to copy this source image file into `target/dita/regions/Content/Images`

IanMayo:
> @IanMayo, I have successfully generated the `regions.dita` file from the `html` file. Can you please provide me with some guidance for the final step? Do I need to create folders in the `regions` directory based on the `alt` attribute of the country names?

hopefully the inline feedback answers your questions

target/dita/regions/regions.dita
ebenezer-melkamu: ok, what about this step https://github.com/DeepBlueCLtd/LegacyMan/blob/96336c5408c30edde35010c0a1a69c8c481c94f0/parser/parser.py#L39

IanMayo:
Note: I want our process to build up. So, we should still perform the previous steps. 

(the script is currently failing if `target` is missing, but that's just because the early steps aren't being run)

IanMayo:
There are some inconsistencies in the generated DITA.  I'll push them in a commit - so you can see in the `delta` what needs to be fixed.

target/dita/regions/regions.dita
IanMayo: Aah, it could be case sensitive.  Let's keep it all lower-case.

target/dita/regions/regions.dita
```diff
@@ -0,0 +1,40 @@
...
+   <area>
+    <shape>poly</shape>
+    <coords>
+     540,326,611,308,681,326,711,286,659,128,618,139,588,199,572,273
+    </coords>
+    <xref format="dita" href="./regions.dita"/>
+   </area>
+   <area>
+    <shape>poly</shape>
+    <coords>
```
IanMayo: actually @AbegaM - it would be easier to read if the `coords` elements were all on one line, too.

target/dita/regions/regions.dita
```diff
@@ -8,41 +8,33 @@
   <imagemap>
    <image href="./Content/Images/world-map.gif"/>
    <area>
-    <shape>
-     poly
-    </shape>
+    <shape>poly</shape>
```
IanMayo: aah, @AbegaM - it's the `prettyfy` that is mangling this.
https://github.com/DeepBlueCLtd/LegacyMan/pull/268/files#diff-a6c6b35454c4bcb3844af5bc2c94deb5fd832767ce4eca996361fb50043a4121R105

Let's think about how to resolve it tomorrow.  I _guess_ we could run prettify, then re-process the tree to remove newlines from `<shape>` elements, then write that.  We may end up doing that a lot - so it should be a method that receives the type of element to _Clean-Up_.

ebenezer-melkamu: @IanMayo I have pushed a commit, instead of using `prettify` i have used the `xml.dom.minidom` library to prettify the code, can you run the code and check the `regions.dita`, please.

target/dita/regions/regions.dita
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" ?>
+<!DOCTYPE topic
+  PUBLIC '-//OASIS//DTD DITA Topic//EN'
+  'topic.dtd'>
+<topic id="links_1">
+	<title>Regions</title>
+	<body>
+		<imagemap>
+			<image href="./Content/Images/world-map.gif"/>
```
IanMayo: the above path needs to be lower-case, please @AbegaM 

The rest is fine :-D

IanMayo (approved):

IanMayo (changes requested):
Aah, could you modify the script so it runs the full process?  Some is currently commented out.

IanMayo:
I have an issue.  The `dita` command is failing. Are we running the command before we have produced the `regions.dita`?

```
ian@Ians-Mac-mini LegacyMan % python -m parser.parser data
LegacyMan parser running, with these arguments: ['data']
Target directory does not exist
target directory created
 [gen-list] [DOTX008E][ERROR] File 'file:/Users/ian/git/LegacyMan/target/dita/regions/regions.dita' does not exist or cannot be loaded.
[move-meta] file:/Users/ian/git/LegacyMan/target/dita/index.ditamap:9:59: [DOTX023W][WARN]: Unable to retrieve navtitle from target: 'regions/regions.dita'. 
[move-meta] file:/Users/ian/git/LegacyMan/target/dita/index.ditamap:9:59: [DOTX027W][WARN]: Unable to retrieve linktext from target: 'regions/regions.dita'. 
     [xslt] file:/Users/ian/git/LegacyMan/target/dita/index.ditamap:9:59: [DOTX009W][WARN]: Could not retrieve a title from 'regions/regions.dita'. Using 'regions/regions.dita' instead. 
Publish complete after 2.601580858230591 seconds. Root file at /target/dita/index.ditamap
The directory target/dita/regions already exists
ian@Ians-Mac-mini LegacyMan % ls target/dita/regions 
```

IanMayo:
> I have an issue. The `dita` command is failing. Are we running the command before we have produced the `regions.dita`?
> 
> ```
> ian@Ians-Mac-mini LegacyMan % python -m parser.parser data
> ```

It took me way too long to spot that bug.  You won't have seen it, since it's only since I'm adding `regions` to the document:
https://github.com/DeepBlueCLtd/LegacyMan/pull/268/files#diff-115820dc304c8940e5dcc42bcda1e9de59c0ddd3a5751ccf0413442abbdcd850

IanMayo (approved):

ebenezer-melkamu:
@IanMayo I couldn't reproduce the error in my end

IanMayo:
> @IanMayo I couldn't reproduce the error in my end

that's ok. 

in a meeting for the next hour. You ok getting started on the next step?

ebenezer-melkamu:
@IanMayo yes i am trying to understand the tasks and the output, i am a little confused but i will figure it out

PR #270, opened 2023-07-27
Add function to process non-standard countries

### LINKED ISSUE #264: Process NS Countries
From the `PD_1.html` file, some links are to countries in the current folder (`Europe.html`), others are to countries accessed from the parent folder.

This second set of countries called `non-standard` countries, and we'll be processing them in this issue.  We take the `Country` name from the `alt` attribute in the world-map imagemap file..

These countries have a non-standard index page, like this one:
https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/data/Britain/Britain1.html

(view it [here](https://deepbluecltd.github.io/LegacyMan/data/Britain/Britain1.html))

It contains a set of image links, each linking to a different category of data.

But, in the real data, these images and their text labels are not associated with each other - they are just put near each other via absolute pixel locations. Grrrr.

As a workaround, each one has been given a new `ImageLinksTable`, that contains the data in an easily accessible format.

In the page will be a custom `ImageLinksTable`.  Convert this table to match the one in [britain.dita](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/britain.dita)

(for now we don't need to worry about related links)

This file should go in `britain/britain.dita`.

Old guidance on how to do this is presented here: https://github.com/DeepBlueCLtd/LegacyMan/issues/237

ebenezer-melkamu:
Fixes #264 

### Modifications
* Add a function to change the `Britain.html` file to `britain.dita` file
* Generted a `britain.dita` file


ebenezer-melkamu:
@IanMayo Please review my Draft PR when you get a chance, the script is generating the `britain.dita` file but i have hardcoded the country name in the script, and i think that is not correct.

IanMayo:
Almost.  With this work we are `walking` the existing document set.

So, the trigger for this process is when we are walking through the `PD_1.html` file, and we encounter a link to a region starting with `../`.  In `PD_1.html` you'll see that both `Britain` and `Iceland` start with `../`, where as `Europe` links to `./europe.html`.

So, countries like Britain (that don't start with a region/continent index file) have the `ImageLinkTable` like `Britain.html`, so we generate the special dita file (like you have here).

So, we won't be hard-coding `britain.html`, but this is how we generate _some_ of the target files for `region.dita`.

IanMayo:
Nearly there.

* [ ] We need to correct the URL in the `regions.dita` file.  Now it can refer to the _real_ `.dita` file for some of the countries.
* [ ] We need to copy the images for the image-map from `data` to `target/dita/regions/$COUNTRY$/content/images`
* [ ] In the new country file (like `britain.dita`) the links are invalid - since their destination files aren't there yet.  So, for now, temporarily specify the current file as the target - like we are for `regions.dita`)
* [ ] This is fiddly, but...  In the real data the ImageLinksTable has two rows. So, instead of looping through the `<a>` elements and putting them all into one row, you'll need an outer loop where you loop through `<tr>` then loop through the `<a>` elements within it.  You'll then create one DITA row for each html row (even though there is only one in the data)

ebenezer-melkamu:
@IanMayo I have copied this `DOCTYPE` from the `dita` output in `britain.dita` and it is throwing an error, do we need a `rich-collection.dtd` file?

https://github.com/DeepBlueCLtd/LegacyMan/blob/d53da18d4259b67db18b27a80794ca435787dd4d/parser/parser.py#L135

IanMayo:
Hello @AbegaM - the `rich-collection.dtd` is in the repo, do we just need a different number of `../` parent steps?

ebenezer-melkamu:
@IanMayo, this is the error that I am getting. The script is trying to access the `rich-collections.did` folder in my `target/dita` directory, and I am able to modify the `href` values in the `regions.dita` file successfully. However, the script is throwing an error similar to this:

Screen Shot 2023-07-28 at 10 45 49 AM

parser/parser.py
```diff
@@ -6,6 +6,7 @@
 import subprocess
 from bs4 import BeautifulSoup
 import xml.dom.minidom
+import re
```
IanMayo: @AbegaM - I don't think this import is used. I think we can delete it.

ebenezer-melkamu:
@IanMayo I will figure out the issue but can you try to reproduce the error on your end when you get a chance, please?

parser/parser.py
ebenezer-melkamu: Ahh, i forgot to remove that, i will do it now

IanMayo:
> @IanMayo, this is the error that I am getting. The script is trying to access the `rich-collections.did` folder in my `target/dita` directory, and I am able to modify the `href` values in the `regions.dita` file successfully. However, the script is throwing an error similar to this:
> 
> Screen Shot 2023-07-28 at 10 45 49 AM

Sure.  There isn't a copy of the file at:
```
/Users/ian/git/LegacyMan/target/dita/dtd/rich-collection.dtd
```

It's at:
```
/Users/ian/git/LegacyMan/dtd/rich-collection.dtd
```

So your relative reference needs to go a bit further back up the file tree.  Or, you need to generate an absolute path reference for the file.

ebenezer-melkamu:
@IanMayo Now i used `"../../../../dtd/rich-collection.dtd"` this path and the `regions.dita` is able to read the `britain/britain.dita` and `iceland/iceland.dita` files and i am no longer getting the error related to `rich-collection.dtd`, but it is throwing other errors when it reads this files, i will figure out this errors, Thank you

IanMayo:
@AbegaM  - I modified the `.dtd` path to prepend `../../`.  

* [ ] The next issue is that `britain.dita` is missing an `id` field. We just need to include `id="britain` in the `rich-collection` at the top.
* [ ] The issue after that is that the href destinations aren't valid. We _temporarily_ have to use the _current-file_ as the link destination
* [ ] The issue after that is that we are missing the target images.  Let's talk through a solution to this _missing-content_ issue. We'll keep on getting it.

ebenezer-melkamu:
@IanMayo I have pushed a change, what i did here is correct right? i have replaced the `href` for the image with the `world-map.gif` file

https://github.com/DeepBlueCLtd/LegacyMan/blob/66bd5d37a3a3c4f3986d361f975f6d6fe6e1c544/target/dita/regions/Iceland/Iceland.dita#L14

IanMayo:
> @IanMayo I have pushed a change, what i did here is correct right? i have replaced the `href` for the image with the `world-map.gif` file
> 

Working perfectly :-D

ebenezer-melkamu:
ok Great, let me know if there is anything i have to modify or fix in this PR.

IanMayo (approved):

PR #273, opened 2023-07-28
Process Category Pages

### LINKED ISSUE #272: Process Category pages
A valuable form of information in the manual is the `classlist`.

It lists all of the `class` instances either under a whole country, or within a category of a country (like `Britain` and `Iceland`).

It contains valuable summary data.  Plus, when there is more detail known, it contains a link to the `class` page.

Here is an example of a category page:
https://deepbluecltd.github.io/LegacyMan/data/Britain_Standalones/Britain_Standalones.html

Here is an example of what we will output to:
https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/legacy/britain_legacy.dita

We know it is a category page because it contains a `<td>` with a `colspan="7"`.  To process this page, we find the `td` with a `colspan` of `7`, then walk back up to find the parent `<table>`.  Then we send that table to `dita`.

As before.  For any links we'll just have to link to the current file.

Note:  so far our strategy has been to read the HTML into a data structure, then loop through that data structure to write the `.dita`.  We are starting to get into files that are more loosely structured, and our HTML parse may fail due to unexpected content.  I'd like us to try this strategy:
- obtain the table content as a `soup`
- steadily convert the `html` elements to `dita` language
- when we convert `<a>` links, be sure to replace the `href` with the href of the current file
- put the new content into the `classlistbody`, and generate the other page content (including flag)

ebenezer-melkamu:
Fixes #272 

### Modifications 
1. Added a `parser_utils` module and isolated util modules from `parser.py` to the new module

* [ ] the country page (britain.dita) will need to be modified to create the correct XREF urls
* [ ] fixes as commented in this commit: https://github.com/DeepBlueCLtd/LegacyMan/commit/b86cc0111be9e56a21d4efa2d847ebbc6b2bb315


IanMayo:
> I'd like us to try this strategy:

Note for @AbegaM .  "if" this alternative strategy works, I think it could sit in the `utils` module, since it's not specific to `LegacyMan` - it will be a general purpose `html` to `dita` converter function.

Note: I did look for an existing html-to-dita converter, but couldn't find one.

ebenezer-melkamu:
@IanMayo There are this `<colspec>` elements in the `.dita` file, should i generate these elements manually using a loop? i don't see this `<colspec>` elements reflected in the `html` file we are parsing

[colespec element](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/4fe9374280c1c40afabb5c9029ef14be10485cf2/dita/regions/britain/legacy/britain_legacy.dita#L13)

IanMayo:
I'm checking if they are essential.

IanMayo:
For now, let's assume the are compulsory.  For this `classlist`, the table will always be 7 columns wide.  So, you could hard-code these colspecs and ids.

IanMayo:
Aah, we do need the specific names, in order for the heading row to occupy all columns:
```
 <entry namest="col1" nameend="col7">Overview of systems in France</entry>
```

ebenezer-melkamu:
@IanMayo after creating the `.dita` files where should we store them, for example i have this `.html` files
  ```
     ../Britain_Complex/Britain_Complex.html
     ../Britain_Standalones/Britain_Standalones.html
     ../Britain_Composite/Britain_Composites.html
     ../Britain_Legacy/Britain_Legacy.html
  ```
Should we store them in this kind of path 
  ```
     dita/regions/britain/britain_complex/britain_complex.dita
     dita/regions/britain/britain_standalones/britain_standalones.dita
  ```

IanMayo:
Yes that's right.

ebenezer-melkamu:
@IanMayo I have generated the category pages, i do need to refine my code, but please run the generated code on your end and let me know what you see

parser/parser.py
```diff
@@ -173,6 +177,97 @@ def process_ns_countries(country_name, link):
...
+
+    #Find a <td> with colspan=7, this indicates that the page is a category page
+    td = soup.find('td', {'colspan': '7'})
+    title = soup.find('h2')
+
+    if td is None:
+       raise ValueError(f"<td> element with colspan 7 not found in the {category_page_link} file")
+
+    #Create the DITA document type declaration string
+    dita_doctype = '<!DOCTYPE classlist SYSTEM "../../../dtd/classlist.dtd">'
```
IanMayo: note: you need a two more '../` levels to get to `classlist.dtd`.

target/dita/regions/Iceland/iceland_complex/iceland_complex.dita
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" ?>
+<!DOCTYPE classlist
+  SYSTEM '../../../dtd/classlist.dtd'>
+<classlist>
+	<title>Iceland - Complex</title>
+	<flag>
+		<fig>
+			<image alt="flag" href="images/iceland_flag.png"/>
```
IanMayo: this line needs to be `../content/images/Flag.png`

ebenezer-melkamu: ok, but when i check the folder `britain/content/images` i see a `Flag.jpg` file there but not `Flag.png` and when i check the `iceland/content/images` folder i do have a `Flag.png` file

IanMayo: I see...  Investigating.

Ok here is the logic to find the URL of the flag:
https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/legacyman_parser/parse_countries.py#L131

So, in the NS-Country page (britain.html), find the `h2`. Then find the next `img` file.  That has the URL of the flag.  Pass that URL as a parameter to the new method, and use that.

ebenezer-melkamu:
Hello @IanMayo, i have pushed a new change, i have fixed the code based on the comments, please run the code from your end and let me know if there is anything i have to fix

IanMayo:
It looks really good. Should be at my desk in 30 mins to double-check.

You should start thinking about the next one, #274  :-D

IanMayo:
One fix. The country pages need to have the hrefs updated, so they point to the new category page, rather than `self`.

Aah, I _expect_ we can now use the correct image too.  The URLs for the image should be in the `./[country-category]/content/images` folder.

So, we will replace the instances of `./britain.dita` and the `world-map.gif` in this snippet:
image

ebenezer-melkamu:
@IanMayo I was having errors when i changed some links, it turns out some of the file names for the images in the `/content/images` folder were uppercase like `FlagS.jpg` but now i have changed everything to lower case and i have also fixed the code based on the above comment

IanMayo:
Hello, just two issues

* [ ] In the category-list page, where the is a link - we don't need the duplicate text label. See `unit_daf` below:
image
* [ ] There can be `a` elements that are anchors, not links.   We only process it as a link if it has an `href`. We don't create a link for a file like [this](https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/data/Britain_Standalones/Britain_Standalones.html#L30).  When we encounter an `a` anchor, we should create a DITA `<anchor>` like [this](https://docs.oasis-open.org/dita/dita/v1.3/os/part1-base/langRef/base/anchor.html) - using `name` as `id`.

IanMayo:
Nice work @AbegaM - now just the `<anchors>`.

ebenezer-melkamu:
@IanMayo I am on it, i will push another change in a minute 👍

ebenezer-melkamu:
@IanMayo I am getting an error saying `"anchor must be declared"`, i think i i have to download a `.dtd` file for the `anchor` element right?

IanMayo:
> @IanMayo I am getting an error saying `"anchor must be declared"`, i think i i have to download a `.dtd` file for the `anchor` element right?

No, let's kick that can down the road :-D   Could you create an issue here in LegacyMan for it please?

I'll merge this issue.

IanMayo (approved):

PR #280, opened 2023-08-01
Add process class method - Part 1

### LINKED ISSUE #274: Process Class files
This task is to convert an HTML page for an installation into a DITA class.

Sample input file:
https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/data/Britain_Standalones/britain_1a.html

Sample output file:
https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/legacy/unit_legacy_a.dita

Note: these elements in the generated `dita` file:
* content goes into `body`, starting with a list of images.  [here](https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/legacyman_parser/parse_images_of_class.py#L41) is code to find relevant images from the page - including the blacklist of images we do not want to process
* two parts of the content come from the table that contains a `<td>` cell with a `colspan` of `6`.   We use that first colspan-6 row as a `divider` for the table. The rows _before_ that colspan become the `summary` block, and the rows after it become the `signatures` block

For the blocks of content extracted, we will ignore the structure, just port it across, converting it to `dita` as necessary.  Note: we'll have to _fix_ any `<a>` anchors in the code - but replacing the href with the current filename.

Elements of source file
* `summary` - first two rows of table with `td` with a `colspan` of `6`
* `signatures` - rows 3 onwards of table with a `td` with `colspan` of `6`
* `propulsion` - contents of a `div` containing an `h1` with word `PROPULSION`
* `remarks` - contents of a `div` containing an `h1` with word `REMARKS`

ebenezer-melkamu:
First part of #274


ebenezer-melkamu:
@IanMayo I have only generated the summary block, i will go ahead and will convert the other `html` elements

IanMayo:
An update, for tomorriow:
- it may have been caused by the merge conflict, but the generated `.dita` was [invalid](https://github.com/DeepBlueCLtd/LegacyMan/pull/280/commits/786d0491b5e9122140507dbdb0015b3ac1fb943c#diff-6e35f09bc495431353faed2561605b5009a0e0da2d3b1c6975fa37e1d5a0e8cb)
- I did some changes to resolve the above in [this commit](https://github.com/DeepBlueCLtd/LegacyMan/pull/280/commits/1db6737eff8fb87988675845922d3d3381629a0b)
- I also did some manual fixes to make the output consistent. I did these by hand though, you'll have to do [these](https://github.com/DeepBlueCLtd/LegacyMan/pull/280/commits/412b4a799f12287a04ead4a1f62b20a3778918a4) changes in your code.  I'll summarise them here:
    - move DTD relative reference up two levels
    - top level element is `class` (with id)
    - insert empty `images` element
    - give id for `summary`
    - give number of columns for summary (we can hard-code it, it's always 6)

ebenezer-melkamu:
@IanMayo could you check the `<signiatures>` block please?

IanMayo (approved):

ebenezer-melkamu:
@IanMayo i am having issues with the `<span>` element, i am parsing the `<p>` and the `<ul>` elements from the `HTML` file and then i am changing the `<ul>` to an `<ol>` element, but i am getting an error indicating i can't use the `<ul>` tag inside `<span>`, is it a must to use a `<span>` element

https://github.com/DeepBlueCLtd/Fi3ldMan/blob/4fe9374280c1c40afabb5c9029ef14be10485cf2/dita/regions/britain/legacy/unit_legacy_a.dita#L79

PR #284, opened 2023-08-02
280 Add code block to parse remarks section - Part 2

ebenezer-melkamu:
Fixes #280 

### Modifications 
1. Modify parser script to parse `Remark` section from the `HTML` file


IanMayo:
Format list content:
```
    <p><ul>
      <li>Mode A</li>
      <li>Mode B</li>
      <li>Mode C</li>
      <li>Mode D</li>
    </ul></p>
```
Gets presented as:
```
			<p>Mode A
Mode B
Mode C
Mode D</p>
			<p/>
```

`<ul>`, `<ol>`, and `<li>` elements can be copied start to `.dita`.

ebenezer-melkamu:
> Format list content:
> 
> ```
>     <p><ul>
>       <li>Mode A</li>
>       <li>Mode B</li>
>       <li>Mode C</li>
>       <li>Mode D</li>
>     </ul></p>
> ```
> 
> Gets presented as:
> 
> ```
> 			<p>Mode A
> Mode B
> Mode C
> Mode D</p>
> 			<p/>
> ```
> 
> `<ul>`, `<ol>`, and `<li>` elements can be copied start to `.dita`.

@IanMayo, is this in reference to the remarks section? Could you please explain it to me?

IanMayo:
> @IanMayo, is this in reference to the remarks section? Could you please explain it to me?

It's in the `Propulsion` section

IanMayo:
The current `content converter` strategy isn't working.  We've ended up changing the sequence of the content.  Here is the original:
image

(image, paragraph, two lists)

Here is the generated content:
image


(list then paragraph)

So, I don't think we can handle the `transform` by cycling through the content, and handling it bit by bit.  I _think_ we need to handle it in this way:
* identify `slug` of html
* take a copy of it (to be sure we can't mangle existing content)
* do `find_all` for one element to change at a time
* do an in-place fix to replace it with new dita-friendly element
* append the modified `slug` to the parent element

Sometimes, it may not be just `swap a for xref`, sometimes we may need to introduce new wrapper elements. But, to do that, we still do `find_all`, then detach the item, insert the wrapper, then re-attach the item.

The above is of value since it doesn't worry about the actual structure - it's just translating it in place.

Note: when we come to `<a>` links in the content we won't just translate it to `xref`, we'll open and translate the file, then insert a link to the changed content.  But, that's for the future :-D

AbegaM:
> The current `content converter` strategy isn't working. We've ended up changing the sequence of the content. Here is the original: image
> 
> (image, paragraph, two lists)
> 
> Here is the generated content: image
> 
> (list then paragraph)
> 
> So, I don't think we can handle the `transform` by cycling through the content, and handling it bit by bit. I _think_ we need to handle it in this way:
> 
> * identify `slug` of html
> * take a copy of it (to be sure we can't mangle existing content)
> * do `find_all` for one element to change at a time
> * do an in-place fix to replace it with new dita-friendly element
> * append the modified `slug` to the parent element
> 
> Sometimes, it may not be just `swap a for xref`, sometimes we may need to introduce new wrapper elements. But, to do that, we still do `find_all`, then detach the item, insert the wrapper, then re-attach the item.
> 
> The above is of value since it doesn't worry about the actual structure - it's just translating it in place.
> 
> Note: when we come to `<a>` links in the content we won't just translate it to `xref`, we'll open and translate the file, then insert a link to the changed content. But, that's for the future :-D

Hello @IanMayo, could you please tell me in which file this code is located in?

PR #303, opened 2023-08-04
Change name of Britain dir + modify parser scripts country_name

### LINKED ISSUE #279: Britain folder does not match name
In the real data, the country folder doesn't always match the text name in the link.

For `Britain`, the data could actually be in `Britain1` (don't know why).

But, in this case, we have to navigate our data from the URLs, not to build up the path.

* [ ] Modify mock data so that `data\Britain` is `data\Britain1`
* [ ] Modify URL for Britain in `data\PlatformData\PD_1.html` to point to content in new folder
* [ ] Fix parser to that it reads content according to link URL, rather than country name

Success for this issue will be the parser producing content for `Britain`.

## Update - walking next level down of tree

I have realised we need to handle this issue at the next level down, too.  In the input data, the folder names don't match the real category names, frequently they are shortened.  So for `Britain_Composites` it may just be `Britain_comp`.  So, when we are reading in the category level data, we (again) need to use the folders from the URLs, rather than generating them.

We can reproduce this in this way:
* [ ] Modify mock data so that `data\Britain_Complex` is `data\Britain_Cmplx`
* [ ] Modify URLs for `Complex` in `data\Britain1\Britain1.html` to use the new folder name
* [ ] Fix parser so that it accesses data using URL instead of proper name

Hey @AbegaM - it's probably worth double-checking that we access the class files (`unit_daf.html`) using the URL rather than the name given to a class.    _Yes_ I've just checked it, and I think we're ok there :-D

ebenezer-melkamu:
Fixes #279 

### Modification
1. Take the `country_name` from the URLs `href` instead of the `alt` attribute
2. Change the name of `Britain` folder to `Britain1`


ebenezer-melkamu:
@IanMayo Note: i have changed the URLs href in the `index.ditmap` file 
 https://github.com/DeepBlueCLtd/LegacyMan/blob/e2aa726c84f63999cc2fbed2ac6eb9efe20e601e/data/index.ditamap#L12

parser/parser.py
```diff
@@ -65,7 +65,8 @@ def process_regions(root_path):
 
         # if link starts with ../ create a dita file for the country => example ../Britain/Britain1.html
         link = area["href"]
-        country_name = area["alt"]
+        #country_name = area["alt"]
+        country_name = os.path.dirname(link[3:])
```
IanMayo: in the next `if` block we test if the link starts with `../`.  If that's true, then it's ok to trim the name like you have here.  But if the link doesn't start with `../` then it could just mangle the real name.

target/dita/index.ditamap
```diff
@@ -9,7 +9,7 @@
     <topicref href="./regions/regions.dita" format="dita">
         <topicmeta/>
     </topicref>
-    <topicref href="./regions/britain/britain_complex/unit_daf.dita" format="dita">
+    <topicref href="./regions/britain1/britain_complex/unit_daf.dita" format="dita">
```
IanMayo: in our _generated_ DITA code I'm happy to use the _real_ country name (that came from the alt tag) for folders, ids, and titles.  This issue is just about how we read in the data.

IanMayo:
Thanks for the update @AbegaM .  Nice work.

I've realised that we also need to handle this issue at the next level down too - in the `category` files.  I've added some explanation to the [issue](https://github.com/DeepBlueCLtd/LegacyMan/issues/279) here.

data/Britain1/Britain1.html
```diff
@@ -32,89 +30,62 @@ <h2>Britain</h2>
...
 <!-- start off with some irrelevant links -->
 <!-- 1. in child of current folder -->
 <div>
-  <a href="./Content/Images/Flag.jpg"
-    ><img src="../Britain_Complex/Content/Images/unit_da.jpg" />
+  <a href="./Content/Images/Flag.jpg"><img src="../Britain_Cmplx/Content/Images/unit_da.jpg" />
   </a>
 </div>
 <!-- missing image child -->
-<div><a href="../Britain_Complex/Britain_Complex.html">Here</a></div>
```
IanMayo, line 83: Hello @AbegaM - please note that we cannot rely on the page name matching the folder name below.  So, could you please make edits so that the page and folder are different?  Maybe the page should be `Britain_Complx1.html`.   You'll have to change the `href` below and the filename ;-)

parser/parser.py
```diff
@@ -348,12 +350,11 @@ def process_class_file(country, class_file_src_path, class_file_target_path, cla
 
         # Prettify the code
         prettified_code = prettify_xml(str(dita_soup))
-
         with open(file_path, "wb") as f:
             f.write(prettified_code.encode("utf-8"))
 
 
-def process_category_pages(country, category_page_link, country_name, country_flag_link, root_path):
+def process_category_pages(country, category_page_link, category, country_name, country_flag_link, root_path):
```
IanMayo: hmm, is the `category` (that came from the link title), or the `folder name` (that came from the URL)?   I think we should support both, since the first on is used for titles on the screen, and the second one is used for parsing the legacy content.
I _think_ `category_folder` is the correct name for the second one.

ebenezer-melkamu: The category is coming from the links title, i am taking it from this line 

https://github.com/DeepBlueCLtd/LegacyMan/blob/53bdf388f72d31b595ef2bc18f3a294f46d0139e/data/Britain/Britain1.html#L36

IanMayo: In that case `category` is the valid title :-)

data/Britain1/Britain1.html
ebenezer-melkamu: @IanMayo Let me know if there is an issue with the new update i pushed

IanMayo (approved):

PR #317, opened 2023-08-23
Remove anchor_id from class file links

### LINKED ISSUE #314: Handle class page link that includes a target
From a category page, we have a link to a class file that includes the target anchor.

So, instead of this:
```
 <td><a href="./britain_1a.html">Unit_1a</a></td>
```

We have this:
```
<td><a href="./britain_2a.html#number2"><strong>Unit_2a</strong></a></td>
```

(note, I've also included the `<strong>` attribute that I've just found in the data.  I hope we're already handling that.

I _think_ this was introduced because the target page wasn't in the usual order, so the anchor takes the visitor to the usual first section.

But, our class files will always have a predictable structure.

So, we can just strip any trailing `#anchor_id` tokens from the end of links that we parse when going through the category table.

When we are working through the table in a `category page`, we can just convert `./britain_2a.html#number2` to `./britain_2a.html` and pass it to the class file parser.

This PR ( #315 ) introduces this pattern in the data.

Note: I've just had a play, and we can clean the class path with:
```
   clean_path = class_file_src_path.split("#")[0]
   process_clas_files( clean_path, ...)
```

ebenezer-melkamu:
Fixes #314 

### Modifications 
* Remove `#anchor_id` values that come after `href` link of class files


parser/parser.py
```diff
@@ -270,12 +270,17 @@ def process_category_pages(
                 # Process class files
                 href = a.get("href")
                 if href is not None:
-                    file_name = os.path.basename(a["href"].replace(".html", ""))
+                    # Remove any #anchor_id value after the href
+                    href = href.split(".html")[0] + ".html"
```
IanMayo, line 274: I found that this worked:
> href = href.split("#")[0]

The above way of doing it seems more complex, but does it solve some problem?

ebenezer-melkamu: Yes it is working fine, isn't it better to remove anything after the `.html` instead of splitting it by using the `#` sign?

IanMayo: I don't know ;-)

Let's go with your way :-D

IanMayo (approved):

PR #327, opened 2023-08-23
Copy country images in the loop

### LINKED ISSUE #314: Handle class page link that includes a target

ebenezer-melkamu:
Fixes #314 

### Modifications 
* Moved the image copy function to the table loop

### DIFF
--- a/parser/parser.py
+++ b/parser/parser.py
@@ -140,6 +140,10 @@ def process_ns_countries(country, country_name, link, root_path):
     dita_table = dita_soup.new_tag("table")
     dita_body = dita_soup.new_tag("body")
 
+    # Create the dir to store the content and the dita files for countries
+    country_path = f"target/dita/regions/{country}"
+    create_directory(country_path)
+
     for tr in img_links_table.find_all("tr"):
         dita_row = dita_soup.new_tag("row")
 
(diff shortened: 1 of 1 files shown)


ebenezer-melkamu:
@IanMayo I have moved the code block which copies the images to the loop, but i am not sure if this fixes the issue that we are having, can you please test it from your end?

ebenezer-melkamu:
@IanMayo As you see in this line what i am doing is pulling all of the images in the `country` folder and copy them to the destination folder, i am not taking the `src` for the images from the `img` element, do you think this is causing the problem?

https://github.com/DeepBlueCLtd/LegacyMan/blob/e5cb93cadf4e12a978d0eda2f819eaae74d26ce4/parser/parser.py#L173

IanMayo:
Aah, he are are looking at the category images so complex, composite, legacy, etc under Britain.

The existing images tend to be in the relevant category image folder.  We need to copy them to the relevant category image folder in our output.

Aaah, ok.  We just need to copy all of the images for the category folders, too.

IanMayo (approved):

PR #335, opened 2023-08-24
321. Clean space values in the image href

### LINKED ISSUE #321: Clean image file URLs
Lots of image URLs include spaces.

This is particularly an issue when we collate the `images` for a class.

When we encounter one, we should replace spaces with `-` characters.

So, we rename the file as we copy it to the target folders, AND we update the `href` to access it.

ebenezer-melkamu:
Fixes #321 

### Modifications 

1. Added a function to replace space chars with `%20`


IanMayo:
Nice work @AbegaM 

But, how about committing a modified image from `data` sub-folder?   Once the transform is complete, and the output is generated in `target/html`, we should see a corresponding change in the `dita` file that referred to that image URL (that now includes one or more %20 chars.

ebenezer-melkamu:
@IanMayo I have modified the `daf_1.jpeg` image to `daf 1.jpeg` for testing, and i have modified the `Britain.html` file and its all working now, please let me know when you can test it

IanMayo:
> @IanMayo I have modified the `daf_1.jpeg` image to `daf 1.jpeg` for testing, and i have modified the `Britain.html` file and its all working now, please let me know when you can test it

That's perfect.  We can see the modified `DITA` page, and then the modified html output.

IanMayo (approved):

PR #349, opened 2023-08-25
347 uppercase file suffix

### LINKED ISSUE #347: Convert file suffixes to lower case
Many of the images have upper-case suffixes: `some_image.JPG`.

DITA doesn't allow this. **Update:** Oxygen on target network reports this as error.

So:
* [ ] @IanMayo to get the reported error, and research cause
* [ ] Rename 'daf 1b.jpeg' to 'daf 1b.JPEG' observe that DITA transform fails (note: may only fail on target network)
* [ ] create function that receives a filename, and returns the filename with a lower case suffix (just suffix, first part of filename remains unchanged)
* [ ] when we copy images from source to target, process each image using the above filename fixer (note: we may wish to  _catch_ the special step where we copy the category image over for the `ImageMapTable`.
* [ ] when we write the `href` for an image, convert the href using the above filename fixer

IanMayo:
Fixes #347 

**note:** initially this just contains the relevant mock data


IanMayo:
It appears Oxygen is no longer failing with an uppercase image suffix.  

Keep this as test data, in case it comes back.

PR #350, opened 2023-08-25
346 join components with POSIX path separator

### LINKED ISSUE #346: URLs in DITA should use `/` as path separator, not `\`
We construct paths using `os.path.join`.  But, on MS-Windows systems this uses the `\` separator.

DITA doesn't like that.  So we need our own path-join command, which always uses `/`

IanMayo:
Fixes #346


IanMayo:
Tested on target environment

PR #353, opened 2023-08-29
Add a td to entry converter for htmlToDITA

### LINKED ISSUE #292: Handle html content within summary and signature tables
We are processing the summary and signature tables.

But, we are not supporting complex content within the cells (such as links, lists, etc).

So, when we are looping through the above tables, call our `htmlToDita` method for the contents of each cell, before the cells get filled in.

This is _probably_ just a change in how we fill in the contents for `dita_entry` here:

Image

Instead of doing `dita_entry.string = td.text.strip()` we do `dita_entry.string = htmlToDita(td.text)` (or something like that)

ebenezer-melkamu:
Fixes #292 

### Modifications 

1. Added a feature to change html `td` to dita `entry`


ebenezer-melkamu:
@IanMayo I have pushed a commit, and the code is now throwing an error saying `colspan is required for entry element` but i have added a `colspan` attribute for the `<entry>` element, does our XML schema supports `colspan` attribute for `entry` elements ?

parser/parser.py
```diff
@@ -16,6 +16,7 @@
 )
 
 from class_files import process_class_files
+from html_to_dita import htmlToDITA
```
IanMayo: we can delete this import, please @AbegaM

parser/html_to_dita.py
```diff
@@ -167,23 +167,45 @@ def htmlToDITA(file_name, soup_in, dita_soup):
...
+    for span in soup.find_all("span"):
+        if span.has_attr("style"):
+            if "color:#F00" in span["style"]:
+                span.name = "ph"
+                span["outputclass"] = "red"
+                del span["style"]
 
     # for ul in soup.find_all('ul'):
     #     ul.name = 'ol'
+    for td in soup.find_all("td"):
```
IanMayo: hmm, I think this processing should still be `process_signatures` block.

ebenezer-melkamu: Yes, but in the `htmlToDita` function i already have a block to change the name of the `td` to `entry`
```js
   if soup.name == "td":
        soup.name = "entry"
        soup['colspan'] = "2"
```

but this code is not changing some of the `td` elements to `entry` and the reason is because there are nested `td` elements like this 

```html
 <td> 
   <td> x </td>
  <td> y </td>
</td>
```

and the `if soup.name == "td"` code is not chaning the inner `td` elements, that is the reason i added the `for loop` to the `htmlToDita` function

IanMayo (approved):
Working

Issue #289, opened 2023-08-03
Process files linked from the current file


IanMayo:
Our pages include link to related files, and sometimes to existing class files that we have already processed.

So, when we processing `htmlToDITA`, perform the following:
- look for `<a href="">link</a>` hyper links
- generate the name for the DITA version of the target HTML file
- check if that file exists. If it doesn't:
    - use "colspan:6" check to establish if it's a class page. If it is, just generate link - we'll let the class processing handle it.
    - if it's not a class file, then:
        - generate target DITA file (as `topic` type)
        - find the _content_ sections of source file
        - for each section:
            - find first heading (h1 through h5 if necessary)
            - translate content to DITA
            - push `section` to body of target file, including first heading as title
        - write file to target dita-name 
- create `<xref>` element in dita, pointing to generated file

### Finding _content_ section of target file

**Update:** `Britain_2A` files for `phase_a_xx.html` use the correct format.

The units to be processed all have this structure:
```
<div id="BottomLayer">
   <div id="PageLayer">
     <h2>Some title</h3>
     <p>Assorted content in here</p>
   </div>
</div>
<div id="BottomLayer3">
   <div id="PageLayer3">
     <h3>Some other title</h3>
     <p>More content in here</p>
   </div>
</div>
```

So, it's `PageLayerxxx` that we are looking for.  See how-to below.

```
dates = soup.findAll("div", {"id" : lambda L: L and L.startswith('PageLayer')})
```

ebenezer-melkamu:
@IanMayo I have a function in my `class_file.py` named  `parse_propulsion`, what if i first find the `href` of the other file and pass the `Propulsion` block to the `parse_propulsion` function?

IanMayo:
There is logic in that suggestion.  I did consider doing it that way.  But, the linked file has a number of blocks of text, other than the `propulsion` block. That same propulsion block is also used for some other similar `class` files.
So, it makes sense to convert all of the other file, and link to it.
Note 1: we do this processing if the `parse_propulsion` fails to find the relevant named block - so it replaces the `else` in `parse_propulsion`.
Note 2: there will be other occasions when we wish to convert a set of `PageLayer` elements in `html` to `dita` (such as when we encounter an `<a>` in `htmlToDita`).  So the code to convert the file should be in a re-usable method.

Here is what I imagine this new `parseNonClassFile` method to look like:
```
    def parseNonClassFile(file_path):
        ' strip anchor off path
[...]

PR #364, opened 2023-08-31
Add a function to parse related_pages

### LINKED ISSUE #150: Class data spread across multiple files

ebenezer-melkamu:
Fixes #150 
Supports #289 

### Modifications
1. Added a function to parse related pages


ebenezer-melkamu:
@IanMayo This PR is to see your point of view, what if i call the `parse_propulsion` function to parse the `Propulsion` block of the related page inside the `parse_non_class_file` function?

parser/class_files.py
```diff
@@ -242,4 +243,33 @@ def parse_remarks(tag, target, dita_soup, options):
         print(f"{options['file_path']} does not have a div element with h1 named REMARKS")
 
 
+def parse_non_class_file(tag, file_path):
+    #Check if there is an html <div id="QuickLinksTable"> </div>
```
IanMayo: Hello @AbegaM - this next block of text should be outside this function.

We have assorted logic to find file to handle (propulsion, remarks, others), either from QuickLinksTable or just an `<a>` in html content.

This method should just receive a file - and it will then convert it.

The external logic will then decide what to do with the href for the converted file.  It _may_ put it into `PropulsionRef` or it _may_ put it into `related-pages`, or it _may_ just include it as the `href` for content in the middle of a paragraph of html.

ebenezer-melkamu:
@IanMayo Could you please review my code, i saw in the task description you put a [reference file](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/br_power_units.dita), should i generate new dita elements like `reference`, `refbody` and so on?

IanMayo:
> @IanMayo Could you please review my code, i saw in the task description you put a [reference file](https://github.com/DeepBlueCLtd/Fi3ldMan/blob/main/dita/regions/britain/br_power_units.dita), should i generate new dita elements like `reference`, `refbody` and so on?

Yes, please :-D

parser/class_files.py
```diff
@@ -219,6 +219,22 @@ def parse_propulsion(tag, target, dita_soup, options):
...
+            current_page_link = os.path.basename(options['file_path'])
+
+            #Remove any #anchor_id from the file link
+            related_page_link = related_page_link.split(".html")[0] + ".html"
+
+            if related_page_link == current_page_link:
+                print('Parsing related page stopped, the link found in the QuickLinksTable is the same as the current page link')
+            else:
+                source_file_path = f"{os.path.dirname(options['file_path'])}/{related_page_link}"
+                parse_non_class_file(source_file_path)
```
IanMayo: @AbegaM - once we have processed the external file, we need to generate a `PropulsionRef` entry pointing to it, like in this file:
https://github.com/DeepBlueCLtd/Fi3ldMan/blob/79cd8f0d22c6e12d9c6aceb8809d122cc46adeea/dita/regions/britain/legacy/unit_legacy_a.dita#L24

IanMayo:
Could you refactor it, please

The `parse_non_class_file` method does not need to know about `propulsion`, it just needs to perform the processing in this block:  https://github.com/DeepBlueCLtd/LegacyMan/issues/289#issuecomment-1699069391

IanMayo:
Ok @AbegaM - have done some tidying, and inserted some [comments](https://github.com/DeepBlueCLtd/LegacyMan/pull/364/files#diff-fbbfd5749f6f81283ced4abafaac2a63e6b4f7ff8bd63863c152022cb6d4f00dR337) to get you started tomorrow.

Aaah, could you also refactor the method into another file please?  `reference_files.py` seems about right.

parser/class_files.py
```diff
@@ -327,83 +327,54 @@ def parse_non_class_file(file_path, title, options):
...
+        dita_reference.append(dita_title)
         dita_ref_body = dita_soup.new_tag("refbody")
 
-        # Add title for the propulsion block
-        dita_title = dita_soup.new_tag("title")
-        dita_title.string = "Reference Propulsion"
+        for page in html_soup.find_all("div"):
+            if page.has_attr("id") and "PageLayer" in page["id"]:
+                print(f"Processing {file_path}")
+                # find the first heading
```
IanMayo: Hello @AbegaM  - we have to loop down through the headings until we find one.  So, I suspect we'll use a `for` loop to go from 1 to 5, then append the index to "h" to use in the `find` call.
The loop will exit early once a heading is found.

ebenezer-melkamu:
@IanMayo When parsing the related pages for the `Remarks`, should the file point to `propulsionRef`? or do we have an element named `remarksRef`?

IanMayo:
I'll have to create a new `RemarksRef`.  I can't do that right now.  For now, let's create a `Remarks` block that contains a `title` and a  `span` - with the span containing an `xref` to the converted content.

IanMayo (approved):

IanMayo (approved):

PR #374, opened 2023-09-01
Refactor parse image function

### LINKED ISSUE #221: improved image scraping
Some classes start off with a block of 3 images.  We wish to include them in the `imagelist` class component.

We're currently scraping images from the whole page of content.

Instead, we should learn to recognise the block that's contains the images, and just scrape them from there.

The images are in divs that sit alongside the div containing the colspan:6 table:
Screenshot 2023-08-30 at 16 43 23

So, we handle the collation of the set of images in this way:
1. Find the colspan:6 table
2. Find it's parent (`div id="Table"` above)
3. Find that element's parent (`div id="PageLayer"` above)
4. Loop through the children
5. Check this isn't the colspan:6 table
6. Check if the element `id` attribute begins with `image`, and it contains an `img` child element.
7. Get the `src` path
8. Get the filename
9. Check the filename isn't blacklisted
10. Create an `image` 
11. Append to the `images` collection.

Reproduce in mock.  **Done: in `unit_daf.html`**.  Mock data introduced in [this PR](https://github.com/DeepBlueCLtd/LegacyMan/pull/363)

Note: for the test data in `unit_daf`, check we don't store `prev_db.gif`.

ebenezer-melkamu:
Fixes #221 

### Modifications
1. Refactor the parse_image function


parser/class_files.py
```diff
@@ -81,22 +82,35 @@ def parse_images(tag, target, dita_soup):
...
+
+    #Find the colspan:6 table
+    td = tag.find("td", {"colspan": "6"})
+
+    #Find the parent table (div id="Table")
+    if td is not None:
+        parent_table = td.find_parent("div", id="Table")
+
+        #Find the parent element of the table
+        page_layer = parent_table.find_parent("div", id="PageLayer")
```
IanMayo: sadly I don't think we can rely on the id. We just find the immediate parent `div`, please.

parser/class_files.py
```diff
@@ -81,22 +82,35 @@ def parse_images(tag, target, dita_soup):
...
+
+        for child in page_layer.children:
+            #Find the img child
+            img = child.find("img")
```
IanMayo: we need to find all of the imgs, and loop through them, please

parser/class_files.py
ebenezer-melkamu: @IanMayo which `id`? is it the `id=Table` or `id=PageLayer"`?

IanMayo: For line 91 we do need to provide the `id`.

For line 94, I think we can just use `parent_table.parent`.  That matches the patterns I see in the data.

parser/class_files.py
```diff
@@ -86,31 +86,30 @@ def parse_images(tag, target, dita_soup):
...
         #Find the parent element of the table
-        page_layer = parent_table.find_parent("div", id="PageLayer")
+        parent_div = parent_table.parent("div")
 
-        for child in page_layer.children:
-            #Find the img child
-            img = child.find("img")
-            if  type(child) is bs4.element.Tag and img is not None:
+        for div in parent_div:
+            img = div.find("img")
```
IanMayo: I know this works for the current data, but we need to allow for multiple `img` in the `div`.  So, this needs to be a `find_all` -> and loop through matches.

I'll modify the data in this branch to introduce an example.

**UPDATE:** hold on, there's a better solution.  I'm happy for us to report an error if more than one image is in there.  So, we still have the current `find` processing.  I'll make the change in your branch :-D

IanMayo (approved):
Tested and working (against mock data)

PR #380, opened 2023-09-04
376 keep current folder names

### LINKED ISSUE #376: Don't rename target folders
We're tidying up the folder structure, by using regions/countries, then categories within the region.

But, where pages link to pages in other folders in the tree we can't use those linkages, since the folders have different names.  This _Could_ be worked around (via lookup tables) it's a leap in complexity that I'd like to avoid. So, let's stick with the current folder structure.

_In the future_ it would be good to tidy the folder structure (based upon attributes in the data).  But, for now, retain the existing folder names (including case).

The outcome of this will be a successful `DITA` parse where the folder names and structure match that in the mock `data` sources.

* [x] Ian to introduce inter-folder link in mock data, to check inter-folder links work.  New inter-folder link generated [here](https://github.com/DeepBlueCLtd/LegacyMan/pull/380/files#diff-ab473c3b393043c89fa749ebf4818083f7852124fcf8a02216ffc68a6aa56f55R38)
* [ ] When we generate links for a category page, don't use the new `country/category` folder structure. Stick with the existing folder structure
* [ ] When going through the links in a category page, do not do `generate reference page` for links beginning with `../` - they will be handled separate.  So, insert the link, but don't generate the page

Note: @IanMayo has created a new link to a class in another category (in `britain_legacy` category page), but it's breaking the export.  That's ok. The work in this issue should avoid that error.

New PR created for this issue here: https://github.com/DeepBlueCLtd/LegacyMan/pull/380

IanMayo:
Fixes #376


IanMayo:
I see we are still using the "tidy" country, not the "original" country for the output folder:
image

image

(`Britain1` vs `Britain`)

PR #383, opened 2023-09-05
Add function to generate file meta data for reference pages

### LINKED ISSUE #381: Bug - passing wrong `options` object
We pass an `options` object to `process_non_class_files`.

It is meant to contain the details of the file currently being handled.

But, at this line we pass the details of the parent file - not the one about to be parsed:
https://github.com/DeepBlueCLtd/LegacyMan/blob/develop/parser/class_files.py#L305

This isn't crashing our code, but it may give unpredictable behaviour.  To solve a particular bug, I'm checking the current filename before outputting some diagnostic code.  But, the `file_name` in the `options` contains the name of a file-parent, not the current file.

ebenezer-melkamu:
Fixes #381 

### Modifications 
1. Add a function to generate file meta data for related pages


IanMayo (approved):
tested and works

Issue #391, opened 2023-09-12
Code review from Robin Wilson


robintw:
As requested by @IanMayo, here is an initial code review of the code in `parser/`. I've included a lot of references to specific lines in specific files - hopefully Github will make them into nice links, but if not then it should be relatively easy to find them using VS Code (Ctrl-G triggers the Go To Line functionality).

Please feel free to ask for any clarification on any of these, or express disagreement (a lot of these are open to interpretation and I am quite happy to discuss them).

## General notes
- **Overall assessment:** In general, the code is well-organised and well structured. Functions are often quite long, but make sense as a conceptual unit so I don't think they should be split. There are some issues with repeated code and use of Python-specific features as detailed below, and path manipulation is a particular concern. Some comments don't match the code (or vice versa!). The `htmlToDITA` function is particularly well structured and documented - some would say it should be replaced with a series of functions to do each transformation, which are called in turn on the soup, but I don't think that would actually make it any clearer or easier to maintain.
- In general, proper path joining functions should be used to join paths, rather than f-strings or other manual string manipulation. Examples include `parser/parser.py:23`, `parser/parser.py:29` and many more. You might want to investigate [PathLib](https://docs.python.org/3/library/pathlib.html) which lets you do nice things like `base_path / dir1 / dir2` and it will do the path joining for you. I think that could significantly clean up various parts of the codebase. It can also replace various calls like `os.path.dirname` etc.
- Docstrings for most functions would be useful
- There are various calls to the `replace_characters` function to get rid of `%20` (eg. `parser/class_files.py:114`). I assume you're doing this to remove URL-encoded spaces from the filenames. That makes sense, but might there be other characters that are URL encoded that need updating? You could use a function like urllib.parse.unquote to remove all of them and then just remove the spaces? Might be more robust - but not essential. **Need input from original author**
- ~~A useful new function could be something like `write_prettified_xml` which takes a dita_soup variable and does `parser/reference_files.py:76` to `parser/reference_files.py:79` returning the filename - this could be used in various places and would reduce a bit of duplication.~~
- ~~`parser/parser_utils.py:83`: Not sure why you've got a definition for `__all__` here or in various other utility files - that's normally just in an `__init__.py`, and importing should work automatically without `__all__` in a normal Python file.~~

## Specific notes
-  ~~`lxml` needs to be added to `requirements.txt`~~
[...]

PR #394, opened 2023-09-14
Fix circular import error

robintw:
Fixes the circular import error reported in https://github.com/DeepBlueCLtd/LegacyMan/issues/391#issuecomment-1719812652.


IanMayo (approved):
Tested, and working.

PR #421, opened 2023-10-05
Generic file processor - second part

robintw:
In progress


parser/html_to_dita.py
```diff
@@ -111,6 +111,8 @@ def htmlToDITA(soup_in, dita_soup, div_replacement="span", wrap_strings=False):
                 # TODO: examine use of centre-aligned DIVs. Do we need to reproduce that formatting?
                 del div["align"]
                 del div["style"]
+        if div.get("id") == "":
```
IanMayo, line 118: @robintw - is this the fix to a specific issue?  There was a div yesterday that mistakenly had an empty string as id (abbreviations).  But, I don't know if there are any more in the data that still need to be fixed.  Could we please also log a warning when we delete content like this?  (since it could be destructive without us noticing it).

IanMayo:
Thanks @robintw - could you also update the `README` to include the new parameter please?

parser/parser.py
```diff
@@ -430,25 +431,23 @@ def process_generic_file_content(self, html_soup, input_file_path, quicklinks):
...
+                                continue
                             div_id = div.get("id")
                             image_tags = div.find_all("img")
                             if div_id is not None and "PicLayer" in div_id:
                                 continue
-                            # print(
-                            #     f"top_value = {top_value}, div_id = {div_id}, n_image_tags = {len(image_tags)}"
-                            # )
-                            # print(div.get_text().strip())
+                            print(
```
IanMayo: should this be printed via a `logging` command?

IanMayo: Aah, from the later print statements I guess this code is _in-progress_.

parser/html_to_dita.py
robintw: Yeah this is to deal with that div in Abbreviations. We won't actually be deleting the div itself, we're just deleting the id value, so I wasn't sure if that counted as deleting content?

IanMayo: aah yeah - we're deleting the id, not the whole element.  This fix is fine :-)

IanMayo (approved):
Working satisfactorily on mock content. Merging to give stable baseline for onsite testing.

PR #440, opened 2023-10-08
Generic file processor - attempt 3

robintw:
In progress

### DIFF
--- a/parser/parser.py
+++ b/parser/parser.py
@@ -436,7 +436,12 @@ class Parser:
                     dita_title.string = h2.get_text()
                     break
             else:
-                logging.warning(f"Could not extract title from {input_file_path}")
+                # If that also fails, then take the first h1 and use that, giving an error if there aren't any
+                h1s = html_soup.find_all("h1")
+                if len(h1s) >= 1:
+                    dita_title.string = h1s[0].get_text()
+                else:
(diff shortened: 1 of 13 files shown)


IanMayo:
Hello Robin, I was a bit nervous of changing the filenames for images, for fear of broken links. I "did" only aim to fix the path for DITA files that we generate, but I realise they still receive incoming links.
So yes, let's fix filenames for files we generate as well as files we copy over, and rely on the DITA publish process to spot any holes in our logic.

robintw:
Done in fe0fd97

parser/parser.py
```diff
@@ -553,15 +553,34 @@ def process_generic_file_content(self, html_soup, input_file_path, quicklinks):
                             # print(f"top_value = {top_value}")
                             if top_value > enclosing_div_top_value:
                                 # Check that the difference isn't too big
-                                if top_value - enclosing_div_top_value < 500:
+                                if top_value - enclosing_div_top_value < 1000:
```
IanMayo: I'm a little nervous about going this large - since we're approaching the vertical separation between separate pages.  Could we try with 800 first please?

robintw: Sure, just tested with 800 and it works. Will commit it now.

PR #450, opened 2023-10-12
Generic file processor - attempt 5

robintw:
In progress

### DIFF
--- a/parser/parser_utils.py
+++ b/parser/parser_utils.py
@@ -215,7 +215,10 @@ def sanitise_filename(filename, remove_extension=False):
         path_obj = Path(basename)
         basename = str(path_obj.with_suffix(path_obj.suffix.lower()))
 
-    return str(p.with_name(basename))
+    new_full_path = str(p.with_name(basename))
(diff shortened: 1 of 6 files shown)


parser/parser_utils.py
```diff
@@ -215,7 +215,10 @@ def sanitise_filename(filename, remove_extension=False):
         path_obj = Path(basename)
         basename = str(path_obj.with_suffix(path_obj.suffix.lower()))
 
-    return str(p.with_name(basename))
+    new_full_path = str(p.with_name(basename))
+    new_full_path = new_full_path.replace("\\", "//")
+
```
IanMayo, line 222: @Robin - one "fix" has a single forward slash, and the other has a double-slash. Is this right?

robintw: No, that's a mistake. I'll fix it and push a change.

robintw: Fixed. It wouldn't actually have caused an error because double forward slashes are normalised to single forward slashes by the Pathlib code, but it's nice to have it right.

IanMayo (approved):

PR #459, opened 2023-10-13
Generic file processor - attempt 5

robintw:
In progress


parser/html_to_dita.py
```diff
@@ -57,12 +58,27 @@ def htmlToDITA(soup_in, dita_soup, topic_id, div_replacement="span", wrap_string
 
     """
 
+    # If this is just a string of content (ie. not a set of tags) then just pass the text through
+    # unchanged
+    if type(soup_in) is bs4.NavigableString:
+        return soup_in.get_text()
```
IanMayo, line 64: I'm getting an error here @Robin.

`NavigableString` object has no attribute `get_text`.

I'm experimenting with just `return soup_in`

PR #480, opened 2023-10-20
Add git script

robintw:
Add script to commit the output in the target folder to a separate 'docs' branch. Just running `./commit-docs.sh` at the command-line should work. Beware that it will give a warning if there are any changes in the working directory in any folder other than the `target` folder.


IanMayo (approved):
looks great :-)

IanMayo (approved):

PR #521, opened 2023-11-09
Develop 88 dita ot

IanMayo:
Introduce DITA-OT package, so we can customise it.


IanMayo:
Close. We don't need this package right now, editing the publishing template should be enough.

PR #534, opened 2023-11-13
Added configurable functionality

### LINKED ISSUE #89: Configurable security warning
We have a security warning banner on each page (#88)

We'd like the background color and text label to be configurable from the Oxygen `Configure Transformation Scenario` dialog.

frank-zimmermann:
Created 3 parameters in the project file, now it is possible to:
- toggle security header and footer between **yes** and **no**
- set the security text
- set the header and footer background color

It is then also possible to set the values in the transformation scenario.

Fixes https://github.com/DeepBlueCLtd/Fi3ldMan/issues/89

### DIFF
--- a/template/F13ldMan/f13ldMan.opt
+++ b/template/F13ldMan/f13ldMan.opt
@@ -11,6 +11,9 @@
             <page-layout-file page="index-terms" file="page-templates/wt_terms.html"/>
         </html-page-layout-files>
         <parameters>
+            <parameter name="webhelp.show.commercial" value="yes"/>
+            <parameter name="webhelp.commercial.text" value="COMMERCIALLY SENSITIVE"/>
+            <parameter name="webhelp.commercial.background.color" value=""/>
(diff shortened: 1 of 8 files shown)


IanMayo:
Nice work @frank-zimmermann.

The next step is to include the headers and footers on the other pages.

A reminder. This is my preferred order of supplementing the template:
image

But, if XSL is the only option I'm happy to go with that.

IanMayo (commented):
Hello @frank-zimmermann , since the two custom xsl files are almost identical (only whitespace differences), is it an option to just use the one file in both extensions?

frank-zimmermann:
> Hello @frank-zimmermann , since the two custom xsl files are almost identical (only whitespace differences), is it an option to just use the one file in both extensions?

No, we need to have this in both files. 
image
Both extension points, `xslt/customMainPage.xsl` and `xslt/customTopicPage.xsl` are needed. But both use the same header and footer information. So, we are very flexible in adapting the code by using won footer and header templates.

IanMayo (approved):

PR #536, opened 2023-11-15
Move search bar into header #95

frank-zimmermann:
The search bar is implemented for all available types of pages:
* topic pages
* index/main page
* search page

@IanMayo, I mentioned that the toggle function, opening and closing the navigation bar was not working on the smaller window.
But the needed <div> was commented out in the HTML example. Now, the search field is always available. I like that, but if this also needs to be hidden on a smaller window, then some layout/CSS adjustment needs to be done.


frank-zimmermann:
I meant the needed `<div>`

IanMayo (approved):

PR #541, opened 2023-11-17
Suppress link hover effect on current page

frank-zimmermann:
I created the folder `page-templates-fragments/libraries`. The folder is also available as the original in the `com.oxygenxml.webhelp.responsive` folder in the Oxygen folder.

The JS file is created in the `resources` folder and will be copied when creating the HTML.

### DIFF
--- a/template/F13ldMan/xslt/customTopicPage.xsl
+++ b/template/F13ldMan/xslt/customTopicPage.xsl
@@ -4,5 +4,49 @@
     <xsl:import href="inc/customHeader.xsl"/>
     <xsl:import href="inc/customFooter.xsl"/>
     <xsl:import href="inc/customSearch.xsl"/>
-
+    
+    <xsl:template match="whc:page_libraries[@page]" mode="copy_template">
+        <xsl:param name="template_base_uri" tunnel="yes"/>
+        <xsl:param name="i18n_context" tunnel="yes" as="element()*"/>
+        <!-- [DOT_DIR]\plugins\com.oxygenxml.webhelp.responsive -->
+        <xsl:variable name="pageLibRefsDir" select="'../page-templates-fragments/libraries/'"/>
+        <xsl:variable name="pageLibRefsFileName">
+            <xsl:choose>
+                <xsl:when test="@page = 'main'">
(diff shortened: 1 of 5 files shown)


IanMayo:
Thanks @frank-zimmermann - but doesn't the XSL only pull in the `current-page` sources for `topic` files?

IanMayo (approved):
Solution works

frank-zimmermann:
@IanMayo, I need to ask what you want to know about the XSL. I believe I do not understand the question.

template/F13ldMan/xslt/customTopicPage.xsl
```diff
@@ -4,5 +4,49 @@
...
     <xsl:import href="inc/customSearch.xsl"/>
-
+    
+    <xsl:template match="whc:page_libraries[@page]" mode="copy_template">
+        <xsl:param name="template_base_uri" tunnel="yes"/>
+        <xsl:param name="i18n_context" tunnel="yes" as="element()*"/>
+        <!-- [DOT_DIR]\plugins\com.oxygenxml.webhelp.responsive -->
+        <xsl:variable name="pageLibRefsDir" select="'../page-templates-fragments/libraries/'"/>
+        <xsl:variable name="pageLibRefsFileName">
+            <xsl:choose>
```
IanMayo, line 14: Hello @frank-zimmermann - the following construct is a bit like a `switch` statement.  It only looks like our `current-handler` gets added to `topic` pages.  Don't we need to add `current-handler` to the other page libraries too?

frank-zimmermann: As I can see, the current handler is only needed for the topic pages, because for pages like search, nothing is selected in the navigation bar (no separate search link)
And yes, you're right, but I copied the template from the DITA-OT package and modified only the path to the newly created `page-templates-fragments` folder because we need it relatively from the XSLT file.

So I let the template untouched without any huge changes.

IanMayo: Aah yes, I see. We have the links in the header, but none can be current because the header links only refer to topics.

Got it :-D

PR #550, opened 2023-11-21
Add separate CSS

frank-zimmermann:
Link from opt file to new CSS containing all previous modifications from modified oxygen.css.
oxygen.css now contains standard definitions from the oxygen template.


IanMayo:
There's a conflict to solve, please @frank-zimmermann .  I guess it may be worth getting an updated `oxygen.css` and then pulling the content out again.

Note: I've a suspicion there may be one or more edits/fixes in the body of `oxygen.css` rather than in our top section. I think we then just have to compare our oxygen.css with a clean one from Oxygen. (by doing `save as template` from export scenario dialog).

frank-zimmermann:
@IanMayo, I deleted the oxygen.css file and copied the standard file from the template folder. That should solve it?!

IanMayo:
I can look tomorrow morning.

IanMayo:
Fixed in #568

PR #558, opened 2023-11-27
Resolve links when storing list of visited files

robintw:


robintw:
Fixes #556

robintw:
Added some more fixes to solve related issues I came across - basically we just make sure all file paths are fully resolved before we store them anywhere (in the 'already processed' set, in the 'shopping list' etc).

IanMayo (approved):
Works!

PR #598, opened 2023-12-04
Process quicklinks on NS country pages that have ImageLinksTables

### LINKED ISSUE #593: Also need quick links on ImageLinksTable page
Some countries start with an `ImageLinksTable` page.

We need to support QuickLinksTable collections on these pages.

~~(I'll create some mock data, and push to develop)~~ No need, `Britain1.html` has an image links table already.

robintw:
Fixes #593


IanMayo (approved):
Tested and working

Issue #613, opened 2023-12-11
Representative welcome/home page content


IanMayo:
The target content for welcome/home experience has diverged from the mock data.

Update the mock data to be more representative, then we can work on how it is presented/published.

IanMayo:
Done.

PR #615, opened 2023-12-11
613 welcome content

IanMayo:
Supports #613 

Note: this PR introduces a parsing problem.  Our `sanitize_filename` logic fails when trying to sanitize a file that sits in the root folder.


robintw:
I've added a commit here to fix the sanitise_filename issue, and also at the same time I fixed a couple of relative links in Welcome.html that didn't work now it had been moved.

IanMayo:
> I've added a commit here to fix the sanitise_filename issue, and also at the same time I fixed a couple of relative links in Welcome.html that didn't work now it had been moved.

That's perfect - thanks @robintw

PR #620, opened 2023-12-13
Don't put converted files in a 'regions' folder any more

### LINKED ISSUE #618: Need `welcome.dita` to sit at root level
If we imagine the published document being distributed via CD, users will open it, and be directed to open a particular file.

The _standard_ first document to open is `index.html`.  But, the DITA publishing engine will always output the `ditamap` home page to index.html, so that filename is already taken.

In this case, we don't want users to start with the conventional index.html home page, we wish to:
1. open a welcome page that (very) clearly indicates what the document is, the organisation that produced it, and the specific version number
2. then open a warning page that reminds them how the sensitive contents are to be handled, and what to do if they accidentally come across this CD
3. from here it's ok to dive into the content. The _traditional_ way is to go into the `regions` map page, but I'm pretty sure the new way will be to type something into the search box.

To support the abive we're going to direct the users to open `welcome.html` - which will open a cover page.

There is just one issue with the above.  Even though the `welcome.html` source file sits in the root of the `data` folder, it ends up in `regions` folder.

I was thinking we needed some custom processing to not put `welcome.html` into `regions` - but that could make some links difficult.  

On reflection, I think it's right to cancel the introduction of the `regions` folder.  It's a throw-back to when I was trying to intelligently process all data, and it was possible to segregate the content.  Now that we are just walking the tree of content I don't think we should put our generated `.dita` content into `regions` - we should just keep it all at the root level.

This would have the side-effect of leaving `welcome.dita` (and thus `welcome.html`) at the root level.

robintw:
Just put them in the main target/dita folder. Fixes #618

### DIFF
--- a/data/index.ditamap
+++ b/data/index.ditamap
@@ -3,22 +3,22 @@
 <map>
     <title>Field Manual v4 Feb 2022</title>
     <topicmeta/>
-    <topicref href="./regions/PlatformData/PD_1.dita" format="dita">
+    <topicref href="./PlatformData/PD_1.dita" format="dita">
         <topicmeta/>
     </topicref>
-    <topicref href="./regions/Britain_Cmplx/unit_anchors.dita" format="dita"> 
+    <topicref href="./Britain_Cmplx/unit_anchors.dita" format="dita"> 
         <topicmeta/>
     </topicref> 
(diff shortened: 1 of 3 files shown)


IanMayo (approved):
Working perfectly :-D

PR #622, opened 2023-12-13
Expand processing code to append captions to image divs if they are present

### LINKED ISSUE #594: Missing picture labels
A number of the `_pics` files contain text labels.  These captions explain what is in the image, for example - it's a screenshot of a system running at 200rpm on a very cold day.  This significant context explains how the image is to be interpreted.

They are floating beneath the image.  These top level divs tend to have an id beginning with `ClickOnThePic`, `ClickOnThePic2`, etc.  In the example I just looked at, the `top` of the `ClickOnThePic` was at 415, and the `top` of the image div was at 5px.  We appear to be correctly pulling in the `image` divs - we just need to associate them with the `ClickOnThePic` with a `top` immediately afterwards - and insert the div contents as text following the `img` in the div.

There are also lots of blocks saying "Click the picture to return to the previous page".  We aren't processing these, since modern users know how to use the browser `Back` button.

But, these text labels aren't going over - and the content checker is finding these instances.

Reproduce the pattern, and come up with a way of handling them.

Sample content in #601

robintw:
Fixes #594

### DIFF
--- a/parser/lman_parser.py
+++ b/parser/lman_parser.py
@@ -32,6 +32,8 @@ from parser_utils import (
     is_button_id,
     is_skippable_div_id,
     does_image_links_table_exist,
+    get_top_value_for_page,
+    append_caption_if_needed,
 )
 
 FIRST_PAGE_LAYER_MARKER = "##### First Page Layer"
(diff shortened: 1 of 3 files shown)


IanMayo (approved):
Tested and working

PR #623, opened 2023-12-13
Process any subsequent content to the main table in category pages

### LINKED ISSUE #600: Trailing content after category table
At least one category table has several paragraphs of trailing content (in the same parent div).

There isn't another obvious page where I could manually move this content to.

We should parse it, and include it after our converted category table.

robintw:
Fixes #600.

### DIFF
--- a/parser/lman_parser.py
+++ b/parser/lman_parser.py
@@ -487,6 +487,10 @@ class Parser:
 
         dita_section.append(dita_emptytitle)
         dita_section.append(dita_table)
+
+        for el in parent_table.next_siblings:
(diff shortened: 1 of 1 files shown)


parser/lman_parser.py
```diff
@@ -487,6 +487,10 @@ def process_category_pages(
 
         dita_section.append(dita_emptytitle)
         dita_section.append(dita_table)
+
```
IanMayo, line 490: Clever bit of code there @robintw . Nicely done :-)

IanMayo (approved):

Issue #619, opened 2023-12-13
Don't centre .xref .b links


IanMayo:
We have styling that centres bold hyperlinks.

It's causing lots of problems.

The style was introduced to centre the labels for `imageTableLinks` labels.

Instead - give imagetables an `outputclass` (hopefully in python), and just apply the block/centre formatting to links under that class.

IanMayo:
@IanMayo to provide specific formatting for `.xref .b` elements that sit under `ImageLinksTable`

PR #624, opened 2023-12-13
Add output class for ImageLinksTable.

robintw:
The output published HTML now has a table element that looks like this:

```
<table class="table ImageLinksTable">
```

Is this what you were hoping for with #619?


IanMayo (approved):
That's perfect, thanks @robintw

PR #637, opened 2024-01-15
Move floating elements (such as tables and images) to the correct chunk of empty space in the page

### LINKED ISSUE #636: Image/Tables floating in whitespace
The first pattern is a simple one.  (_Update: oops, not it isn't_).

An image is "floated above" the content in a div or span.  But, in the html the image is defined at the correct logical location, followed by a block of whitespan.

We should identify this as follows:
- find a floating div/span at the start of some whitespace
- remove the floating parent element
- delete the successive whitespace

Hmm, actually I don't think the above is a satisfactory solution.  There's a fair chance that image is actually positioned somewhere else on the page, over a different block of whitespace.

So, sadly, I think we have to handle the generic instance:
1. find all floating divs/spans (where they are not blacklisted either by the div name or image filename)
2. walk back up parent tree to generate `top` coord in the whole-page coord space
3. order floating div/span contents by `top`
4. create list of whitespace blocks within page elements of the the file
5. create list of top level elements that contain whitespace, generate the `top` for these parent elements
6. for each top level elements, find images with `top` value >= that element `top`, but < the next element `top`
7. sequentially insert grouped images/tables into whitespace for top level blocks, ensuring the calculated `top` of the image is greater than or equal to the `top` of the closest absolute-positioned parent of the whitespace. 

Note: once images have been inserted into whitespace, other blocks of whitespace should be removed.

A common pattern for the above is demonstrated below.  Note how issue 7 above ensures this image is associated with this whitespace. The coords of the whitespace parent (`1516px + 0px`) is `<=` the calculated top of the image (`1516px` + `0px` + `73px`)
https://github.com/DeepBlueCLtd/LegacyMan/blob/9a5790474ba5493d1253397123211213010c8904/data/Britain_Cmplx/unit_a28.html#L302

Oops, I've got to take this back. We do _almost_ handle the above correctly.  We remove the `floating` parent of the image, and put it inline.  In the above case, the image is left in the correct place, and we just have the trailing whitespace to remove.

Will carry on looking for instances tomorrow (Tues) that do require positioning to be fixed.  I do know there are lots of them - there's a pattern where the images are all declared right at the end of the file.

Aah, actually - the above pattern (image is floated, but inserted into content at the correct logical place) does represent a high proportion of the issues.  We have code to `remove floating parent` for images.  _Maybe_ we modify that to check it's roughly in the correct place (a low `top` coord: under 800px seems to work so far) - and if that's the case we just remove any trailing whitespace.  If we defer `warn-blank-lines` until after the above processing is complete, we should have a lot fewer instances to track.  File `unit_a28` does represent a valid case for the above.

Ok, `France1\FR_A27_Unit.html` includes a floating table and an image, where the image is (slightly) in the wrong place.  Also `Britain_Complex\unit_28.html` includes an image that is in the correct place.  Oh, and there is a floating propulsion table in `Britain.Legacy/unit_delta(13).html`

robintw:
Currently in-progress and uploaded for testing.

Fixes #636.


IanMayo:
Have introduced model pattern of data that gives this error when run from `run_defloat_conversion.ppy`:
> no suitable floating elements for blank space at line 418 of data/Fr_Legacy/FR_100_unit_charlie.html

IanMayo:
Note: I get this pattern hundreds of times:
```
Warning: no suitable floating elements for blank space at line 123 of /Users/ianmayo/git/LegacyMan/data/Spain.Legacy/sp_charlie_pics.html
Warning: no suitable floating elements for blank space at line 201 of /Users/ianmayo/git/LegacyMan/data/Spain.Legacy/sp_charlie_pics.html
Warning: no suitable floating elements for blank space at line 217 of /Users/ianmayo/git/LegacyMan/data/Spain.Legacy/sp_charlie_pics.html
Warning: no suitable floating elements for blank space at line 233 of /Users/ianmayo/git/LegacyMan/data/Spain.Legacy/sp_charlie_pics.html
```

It's for pages of images, and a lot of the content sits inside a top level `PicLayer` element.  The `SP_Charlie_Pics.dita` and the HTML file generated from it are both valid.  So,  I guess our subsequent logic handles it correctly, but it would seem to be useful if the defloat logic handled it correctly, since that may help in other places.

IanMayo:
One last item:
>  [gen-list] file:/Users/ianmayo/git/LegacyMan/target/dita/Spain.Legacy/Bravo_Transducer.dita Line 119:The content of element type "ph" must match "(boolean|cite|keyword|markupname|apiname|option|parmname|cmdname|msgnum|varname|wintitle|numcharref|parameterentity|textentity|xmlatt|xmlelement|xmlnsname|xmlpi|ph|b|i|line-through|overline|sup|sub|tt|u|codeph|synph|filepath|msgph|systemoutput|userinput|menucascade|uicontrol|equation-inline|q|term|abbreviated-form|text|tm|xref|state|data|sort-as|data-about|foreign|svg-container|mathml|unknown|image|draft-comment|fn|indextermref|indexterm|required-cleanup)".

This is related to our extra div elements. The parent one is converted to a `ph`, but a `ph` cannot contain a `p` as in this generated DITA code:
image

I _expect_ our `fix extra divs/spans` will resolve this.

robintw:
Is this issue in your first comment (`Have introduced model pattern of data that gives this error when run from`) the same as the issue in the second comment (`Note: I get this pattern hundreds of times:`)?

The second one occurs because of the way the pics files are written: everything is within a `picLayer`, and so every bit of repeated blank space is counted as one that we might want to fill (as they're inside a div not just inside the base body tag). Should I stop reporting those errors when they appear within a picLayer div? The same logic is probably producing useful warnings for non-picture files.

IanMayo:
> Is this issue in your first comment (Have introduced model pattern of data that gives this error when run from) the same as the issue in the second comment (Note: I get this pattern hundreds of times:)?

No, they are separate observations.  The [new](https://github.com/DeepBlueCLtd/LegacyMan/pull/637/commits/0f7e6cdc5c6effd5f3d843b24f5ed54a65c1e27b) instance of mock data throws a single `no suitable floating element` error.   But, it's a different pattern that throws the multiple `no suitable floating elements` for one file issue.  I think they happen in the `picLayer` content pattern.

Would it be possible for our code to handle the pattern where lots of content is in a top level `picLayer` element? There are lots of files that match that pattern, and it would be reassuring to know we're handling it via deliberate `defloat` logic.  Aaah no.  _Maybe_ our existing functionality is acceptable - where we match floating image elements with the `number2` HTML anchor they relate to (according to the shopping list) .  If that's the case - then we shouldn't raise a `no floating element` warning for those files.  I'm pretty sure we successfully remove the whitespace from those files too (since it's between PageLayer elements, not inside them.

robintw:
I've pushed quite a few updates this evening - sorry for my lack of work on this over the last week (other work got quite busy).

I think I've fixed the issues you raised here with the new example, and stopped reporting errors for pics files. I then spent quite a while battling BeautifulSoup while dealing with unit_banjo.html. I think I've solved it now, but it isn't solved in a very nice way.

So, I have a question:

At line 908 of `data/Wales_Cmplx/unit_banjo.html` there is a `BottomLayer3`, which doesn't have a `PageLayer` child. Instead, there is a `layer1` child. How common is it for pages to _not_ be in a PageLayer child? And if there isn't a PageLayer child, is it always called something like `layer1`?

The problem we're running into here (which I sort of described in a comment by the yucky code, but not very well) is that initially we weren't finding the table in that BottomLayer as a floating element, as it wasn't in a PageLayer. But, once we changed our logic to deal with floating elements that aren't inside PageLayers, we picked up this `layer1` as the floating element that we need to place. Unfortunately the blank elements that we want to replace with the floating element are _inside_ that floating element - so we end up with circular references and chaos ensues. The yucky code currently searches for these circular references and tries to remove them - which seems to work, but is probably quite slow (particularly when run on the real data). If we could just ignore things like `layer1` inside a BottomLayer then that'd be a lot easier (by ignore here, I mean not return `layer1` as a potential floating element to be moved, but just any children of `layer1` that are floating elements).

I hope that vaguely makes sense, it's late and I'm tired.

Let me know what you think.

IanMayo (approved):
Tested and working.  Merging early, to test against target content.

IanMayo:
> At line 908 of data/Wales_Cmplx/unit_banjo.html there is a BottomLayer3, which doesn't have a PageLayer child. Instead, there is a layer1 child. How common is it for pages to not be in a PageLayer child? And if there isn't a PageLayer child, is it always called something like layer1?

Now I've gone to inspect that file in the real content, I see I rename `layer` to `PageLayer3`, since it was holding me back.  I'll look out for any other instances of this pattern.

PR #647, opened 2024-02-09
Extract title from regions page and use as DITA title

### LINKED ISSUE #644: Use correct region title in Region files
`Europe.html` is an example of a `region` file - it lists the countries that appear in that region.

Our current processing is to give each of these a title which says `Regions`, and the convert the rest of the page.

Each region file actually has the region title in a `h1` element.  This would make a better `title` element.

We could support this as follows:
- introduce a flag to `htmlToDITA` which instructs it to strip out `h1` elements
- in `process_sub_region`, find the first `h1` in the file, and use this as the title.  Instruct `htmlToDITA` to remove h1 elements in this case.

robintw:
Fixes #644.

### DIFF
--- a/data/PlatformData/Europe.html
+++ b/data/PlatformData/Europe.html
@@ -5,7 +5,7 @@
 <div style="position: absolute; visibility: visible; left: 164px; top: 226px; width: 455px">
   <table width="111%" height="189" border="1">
     <tr>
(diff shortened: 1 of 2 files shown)


IanMayo (dismissed):
Nicely done, and a lot simpler than my proposed strategy 🤣

IanMayo (approved):

PR #651, opened 2024-02-16
649 layout of flag header

### LINKED ISSUE #649: Update display of flag data
We currently have a `flag-section` in the data.

We have a new content model where we display the flag in the document title, using this structure:
image

We should remove the `flag-section` and update the `title` section to match the above model.

Aah, but for the `country`  page - we don't need to include the link (xref) back to the parent (self) file.

IanMayo:
Fixes #649


IanMayo:
Tested and working on target data

PR #653, opened 2024-02-23
Fix category pages being overwritten

### LINKED ISSUE #652: Country page being over-written with generic version
I thought I'd spotted a pattern in our mock data yesterday, but today I've seen the pattern in the real data.

A stakeholder has reported that the flag is missing from a country. I've checked and re-checked the structure of the flag data, and it matches other flag data.

I've also inserted a debug line, just before the flag is added to the dita, and it's present.

But, after the `publish` is complete, there is no flag.

I'm pretty sure the version of the file produced by the country/category specific parsing is being overwritten by a generic parser.

Here is my comment regarding the same pattern of behaviour in the `Spain` file: https://github.com/DeepBlueCLtd/LegacyMan/pull/646#issuecomment-1946436381

Aah, I think I have a solution.  In our generic file processor, we start with a quick check to establish that it's not a file that is subject to `Special` processing.  We can skip the category files since they have a TD:7. Aah, no. Transducer files also have that. I _guess_ that in phase 1, when we process the category files under the regions, we could add a flag to the dictionary saying "no_generic_processing".

We could check for another phrase. These country/category files all have a TD:7 containing "PROPULSION CHARACTERISTICS OF..." (though our mock data doesn't currently have this).  That could be an easy fix.

Aah, I've just thought of an even simpler solutiuon.  Then generic file processor should not run if the file has been already processed.  Then, if "custom processor" runs that page first it won't get over-written, but if "generic processor" runs first it will get over-written.  That should solve it :-)

**Update:** I've just been investigating this.  We are tracking `files_already_processed`.  So, if we handle `category` processor first, it should not get over-written by `generic` processor.  Aah, ok. For `category processor` the file id stored in `files_already_processed` includes the path, but in `generic processor` it isn't:
```
Category processor: target/dita/Spain/Spain.dita
process_generic_file: Spain.dita
```

But, ensuring we're storing the same path _may_ not resolve this.  It's still possible for a file to be processed by the `generic file processor` first, which (I _think_) would prevent it being processed by the category processor.

robintw:
Ensure that all paths that are added to `files_already_processed` are full, absolute paths rather than being some sort of relative path or just a filename. This ensures that files aren't overwritten by the generic file processor after they've been written by the category file processor.

Fixes #652.

### DIFF
--- a/parser/lman_parser.py
+++ b/parser/lman_parser.py
@@ -587,6 +587,7 @@ class Parser:
 
         category_file_path = f"{category_path}/{os.path.basename(category_page_link)}"
         write_prettified_xml(dita_soup, category_file_path)
+        category_file_path = Path(category_file_path).resolve()
         self.files_already_processed.add(category_file_path)
 
         input_file_path = Path(f"{self.root_path}/{remove_leading_slashes(category_page_link)}")
--- a/parser/lman_parser.py
+++ b/parser/lman_parser.py
@@ -183,7 +183,7 @@ class Parser:
         os.makedirs(regions_path, exist_ok=True)
 
         write_prettified_xml(dita_soup, f"{regions_path}/PlatformData/PD_1.dita")
-        self.files_already_processed.add(f"{regions_path}/PlatformData/PD_1.dita")
+        self.files_already_processed.add(Path(f"{regions_path}/PlatformData/PD_1.dita").resolve())
 
     def process_sub_region(self, link):
         with open(link, "r") as f:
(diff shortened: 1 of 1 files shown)


IanMayo:
This fix works for `Spain` in the mock data. I'll have to test onsite against the target data. I _think_ we may have a problem if a page is calling the `generic` processor before the `category table` processor.

robintw:
Yeah I checked it worked for Spain.

We shouldn't have that problem, because the category processor doesn't check whether the file has been processed already - it just processes it regardless. That should mean that if the generic file processor has already processed it, then it'll just be reprocessed by the category file processor - but once it's been processed by the category file processor the generic file processor _won't_ reprocess it (as the generic file processor _does_ check whether it's been processed already before processing it).

IanMayo:
That sounds perfect @robintw .

IanMayo (approved):
Merging, so it can be tested against target data

PR #657, opened 2024-04-11
640 process flag from page

### LINKED ISSUE #640: Propagate flag (and dimensions) through to class files
When we are looking at a class file, it's easy to forget which nation this is for.

This is particularly significant if/when we are looking at a class with a name that is in use across multiple countries.

It would be useful to prepend the class page title with the country image (and maybe even the country name).

The title block for class files currently looks like this:
image

We can legitimately extend it to include the flag using this markup:
image

As we walk the tree, we should pass the URL for the country's flag image, plus the image dimensions to the `class` file - so that we can prepend the image.  While we are doing this, we should _probably_ also pass the country name through, for if/when we need to display the country name.

**Update:** we should make the flag clickable, to go back to the country page.  To support this we should also pass the URL for the country page.

**Update 2:** I've just been working in this code, and have these observations. The data is already passed to `process_category_pages`.  I _think_ we could add some params to `process_generic_file`.  If that method receives these params it will use them to generate a fancy header block.

The DITA should be structured like this:
image

**Update 3:** We do have a _Special-Case_.  It's for the Non-Standard countries that start with an image table (`Britain`).  These start with an image table, with links to standard category listings.  For these, we need to capture/include the country flag in the non-standard country page (#650), and the pass that image/URL to the category listing page. The flag on the category listing page will then link back to the non-standard page.  So, `Britain` will have a flag/title which doesn't link to anything.  The "complex/legacy.etc" images will link to `britain_legacy.html` - a category page. The flag image on this page will link back to the `Britain` country page.  For class pages that are on the category page, their flag link will also go back to the parent `Britain` country page.  For countries that have a standard category listing, the flag won't link to anything on the category page, but the class pages will link back to the standard categroy listing.

IanMayo:
Fixes #640


IanMayo:
Merging, so I can test on target data.

